PR #53, opened 2015-01-23
Add the failIfUnknown option

ppalaga:
The motivation for this change is to introduce an option that makes it impossible that there are license-header-less files left in the projects.

From the failIfUnknown JavaDoc:

> You can leave this flag on `false` if you do not want the build to fail for files that do not have an implicit or explicit comment style definition. Setting this explicitly to `true` is a safe way to make sure that the effective file type mappings cover all files included from your project.


mathieucarbou:
Thansk ! :+1:

PR #55, opened 2015-01-27
Clarify license of license-maven-plugin (!)

stain:
I assume it's Apache license, as that's what the [parent says](https://github.com/mycila/pom/blob/master/pom.xml#L83).


mathieucarbou:
Yes :-) Thank you!

PR #72, opened 2015-04-14
fixes #70, temporary correction that makes tests to pass even under windows

### LINKED ISSUE #70: tests are failing on windows
On windows some tests are failing as exemple RemoveMojoTest#test_remove_script_style.

The errors are mainly because loaded files contain `\r\n` whereas the test itself checks against `\n`.
Of course changing the tested string to `r\n` corrects the behavior, but why the hell is it working under linux boxes of travis-ci?

McFoggy:
usage of platform line-separator in some tests where it make sense

now with tests running also on windows real re-factoring & contributions can start


mathieucarbou:
Thanks :)

PR #73, opened 2015-04-14
fixes #71 allow to skip multiple consecutive lines

### LINKED ISSUE #71: skipLinePattern: bug or missundertsanding
Hi,

trying to define my own style for some proprietary file formats, I could not skip correctly first lines of my file.
Then I started to debug HeaderParser#findBeginPosition and found the line 
`while (line != null && !headerDefinition.isSkipLine(line))`
why is there a negation before isSkipLine? shouldn't it be the exact opposite, in pseudo code `while ( current line has to be skipped ) read next line`

Did I incorrectly understood the usage of this skipLinePattern value, or is it a real bug?

Problem is easy to reproduce : for example in doc/doc9.xml duplicate the first line and relaunch HeaderParserTest. test_parsing_xml6() will now fail being unable to detect correctly the header because the 2 first lines where not correctly skipped.

McFoggy:


mathieucarbou:
Thanks :-)

PR #75, opened 2015-04-16
introduce SVN plugin

McFoggy:
Mathieu, 

find enclosed a contribution that adds a subversion plugin to add properties for subversion hosted projects.

Matthieu (with 2t ;-) aka McFoggy


mathieucarbou:
Nice! Thanks :-)

PR #89, opened 2015-09-07
Changed the double tab to one tab

nberlijn:
The current Apache License, Version 2.0 has two tabs (8 spaces). But the officiel Apache License, Version 2.0 has one tab (4 spaces); http://www.apache.org/licenses/LICENSE-2.0.html#apply


mathieucarbou:
Thank you!
I think we must find another way to fix this because the provided pull request would have a lot of impacts on all users who use the pre-configured license header.

nberlijn:
You can make a new release (bug-fix), or host it somewhere else. I resolved it for my own projects by using and hosting my [custom template](http://licenses.nbsoft.org/templates/APACHE-2.0). Notice, I use a custom `${year}` attribute.

Issue #76, opened 2015-04-16
enforce project EOL to be correct regardless of user GIT settings


McFoggy:
We should find a way to enforce some rules at project level so that whatever the default global git configuration of a user is, the project works as expected.

Today (commit [dbc060a5297228055ce791ad2af4027da6e37a95](https://github.com/mycila/license-maven-plugin/commit/dbc060a5297228055ce791ad2af4027da6e37a95)), if you are on a windows machine and you applied different GIT recommandations ([GIT SCM book](http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration), [Github documentation](https://help.github.com/articles/dealing-with-line-endings/)) by setting `git config --global core.autocrlf true`
then the project does not build correctly: tests in UpdateMojoTest & RemoveMojoTest are failing because they rely on specific `\n` line ending.

At least we should investigate if we could not enforce, using git functionnalities, project specific rules that would override user settings.

mathieucarbou:
As I already said, `core.autocrlf` should NEVER be true for the simple reason that it is a host-specific config which modify project files.
If prject files are modified, it is a NORMAL behavior for tests to fail.

mathieucarbou:
I'll add that there is a way to modify this property at a project level by editing the file `.git/config`.

McFoggy:
I do not understand why you are closing this directly without letting anyone including me to lookup & provide a way to make this transparent?
From what I see on projects where EOL characters matter, they provide PROJECT SPECIFIC rules via a .gitattributes file.
If there is a way to do this clean, why do you refuse to at least investigate it?

Please reconsider reopening the issue until a real analysis of existing possibilities is done.

McFoggy:
>  I'll add that there is a way to modify this property at a project level by editing the file .git/config

but that means a manual step after checking out the project.
Really I am convinced that if a clean setup is possible then we should try it.

mathieucarbou:
Sorry! yes it should not be closed! I misunderstood the issue description. If here is some way to set the `autocrlf` setting specifically for a project, yes it would do the trick.
I'd like also to avoid placing a `.gitattributes` file which treat all files as binary ;-)

PR #96, opened 2016-02-02
normalize line ending

McFoggy:
this PR answers the line ending problems raised in #76 

As the project, and its tests, need special line ending to be green I do not think that the project can afford to introduce line ending normalization. By doing this normalization and by specifying concrete line ends where it matters the project is build-able and tests pass regardless of the configuration of line ending on the client machine that checkouts the project.

#### The work done:
- normalize line endings in source code, no need to mix CRLF with LF in sources
- use LF as default normalization: as most of the tests check for a LF it was easier to define LF as default and just express on which files CRLF was needed
- identify files for which CRLF was mandatory
- provide `.gitattributes` file answering the above points

#### Usage:
- on a clean working copy: just git clone and you are done (regardless of any `core.autocrlf`settings value)
- on an existing working copy:
  - merge the PR
  - delete everything except `.git` directory
   - on linux using bash you could execute the following commands
```
shopt -s extglob
rm -f !(.git)
rm -r -f !(.git)
git reset --hard
```

Using this version you should be able to have all tests green regardless of any git settings you could use on your machine. This normalization also ensures that even if someone try to push bad line endings where not expected, then those ones will be normalized by the rules in `.gitattributes`.

If requested I can squash the PR. I kept it until now  with several commits to show how I ended in the final state.

_readings_:
- http://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/
- http://git-scm.com/docs/gitattributes


mathieucarbou:
@McFoggy : sorry I just saw that I was too busy... I had a look.
Did you tried a `mvn clean install` on both linux/mac and windows ?

McFoggy:
at the time of the PR, I tried, windows & linux with different `core.autocrlf` settings.

Issue #41, opened 2014-06-09
license:remove is not working


masih:
The `remove` goal does not remove file headers despite the log messages saying the headers are removed.

My plugin configuration: 

``` xml
<plugin>
    <groupId>com.mycila</groupId>
    <artifactId>license-maven-plugin</artifactId>
    <version>2.5</version>
    <configuration>
        <header>com/mycila/maven/plugin/license/templates/GPL-3.txt</header>
        <excludes>
            <exclude>**/README</exclude>
            <exclude>src/test/resources/**</exclude>
            <exclude>src/main/resources/**</exclude>
        </excludes>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>check</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

The command I execute is `mvn license:remove`.

The (shortened) log:
```
[INFO] Scanning for projects...
[INFO] 
[INFO] Using the builder org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder with a thread count of 1
[INFO]                                                                         
[INFO] ------------------------------------------------------------------------
[INFO] Building XXX 1.0-SNAPSHOT
[INFO] ------------------------------------------------------------------------
[INFO] 
[INFO] --- license-maven-plugin:2.5:remove (default-cli) @ XXX ---
[INFO] Removing license headers...
[INFO] Removing license header from: XXXXXX.java
.
.
.
[INFO] Removing license header from: XXXXXX.java

[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 2.379 s
[INFO] Finished at: 2014-06-09T14:11:58+00:00
[INFO] Final Memory: 9M/182M
[INFO] ------------------------------------------------------------------------
```

I have enabled `-X` and no exceptions are being thrown by the plugin.
Any clues?

Many thanks

mathieucarbou:
Please provide a junit test to reproduce. Or provide a pull request.

McFoggy:
I succeed in reproducing the issue under a branch of [my fork](https://github.com/McFoggy/license-maven-plugin/tree/issue-41).

Basically actions to reproduce:
- from a file without header
- using `format` goal add GPL-3.txt header 
- call remove goal

A user could expect that the header is removed, I was such a guy until I discovered the keywords parameter on the mojo.
[...]

PR #101, opened 2016-04-25
Issue #41 - Fix license templates

mathieucarbou:
- (c) and (C) were both converted to ©
- '${project.name}' was replaced w/ 'This program'
- GPL licences got their proper introduction

See https://www.gnu.org/licenses/gpl-3.0.html and https://www.gnu.org/licenses/gpl-howto.html for information about how to use GPL licences.


mathieucarbou:
see #41

PR #114, opened 2016-08-16
Add The MirOS Licence

mirabilos:
This is actually a Free Software licence and OSI certified.

This is part 1; part 2 would be to actually list all relevant copyright years,
and not a range (dash-delimited — actually ‘–’-delimited (en dash),
not ‘-’-delimited (hyphen-minus)) either, but that’s still open:
https://github.com/mycila/license-maven-plugin/issues/95


mathieucarbou:
thanks!

PR #120, opened 2016-09-23
When a multi-line inlineHeader is provided, remove indenting

akbertram:
Ideally, you should be able to include an `inlineHeader` like so:
```.xml
 <plugin>
        <groupId>com.mycila</groupId>
        <version>3.1-SNAPSHOT</version>
        <configuration>
          <inlineHeader>
              Renjin : JVM-based interpreter for the R language for the statistical analysis
              ${project.name}
              Copyright © ${project.inceptionYear} BeDataDriven Groep B.V. and contributors
              
              This program is free software; you can redistribute it and/or modify
              it under the terms of the GNU General Public License as published by
              the Free Software Foundation; either version 2 of the License, or
              (at your option) any later version.
              
              This program is distributed in the hope that it will be useful,
              but WITHOUT ANY WARRANTY; without even the implied warranty of
              MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
              GNU General Public License for more details.
              
              You should have received a copy of the GNU General Public License
              along with this program; if not, a copy is available at
              https://www.gnu.org/licenses/gpl-2.0.txt
          </inlineHeader>
          <mapping>
            <R>SCRIPT_STYLE</R>
          </mapping>
          <includes>
            <include>**/*.java</include>
            <include>**/*.R</include>
          </includes>
        </configuration>
      </plugin>
```
However, in this case the ident from the XML file is replicated across the source files. This PR trims each line if multiple lines are provided.

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderSource.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderSource.java
@@ -91,7 +91,7 @@ public abstract class HeaderSource {
      */
     public static HeaderSource of(String inlineHeader, String headerPath, String encoding, ResourceFinder finder) {
         if (inlineHeader != null && !inlineHeader.isEmpty()) {
-            return new LiteralHeaderSource(inlineHeader);
+            return new LiteralHeaderSource(reformatInlineHeader(inlineHeader));
         } else if (headerPath == null) {
             throw new IllegalArgumentException("Either inlineHeader or header path need to be specified");
         } else {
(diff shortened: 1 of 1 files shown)


akbertram:
@mathieucarbou Any feedback on this PR? Happy to make any changes needed to get it merged!

mathieucarbou:
Sorry i didn't see the notification!

When you say trim, it means all spaces before each lines are removed ?

If so, it would not work with every licenses...

Within an XML tag, the text is kept as is an this is normal. A little like groove's multi-line comments.

If you wish to not see the spaces before each line, you should not put them inside the inline header. The way to see that is like you had a chat section where the whole text content is taken as-is.

So I am not sure about this PR since I guess it would break headers that specifically wants some spaces bedore, to mimic some indents for example.

Does it make sense or I misunderstood something ?

akbertram:
Ah I see what you mean, yes, then it wouldn't be possible to have a header with something indented. There is a workaround for this - we just [avoid indenting](https://github.com/bedatadriven/renjin/blob/master/pom.xml#L169) that part of the XML, so I'm ok if you want to decline this PR.

PR #132, opened 2017-11-25
fix: Use at least a single thread for the checks

rhuss:
If forking is disabled or the cast to an integer number results in 0,
then the ExecutorService wont start up.

Actually, it would be awesome to allow also an absolute number for the
threads to use. The reason is, that `Runtime.getRuntime().getAvailableProcessors()` is not cgroups aware, which means, that even in container restricted environments (Docker, Kubernetes) this number is always the number of cores of the host system.

As actually many CI systems are running in the cloud, possibly
in shared environments on huge machines with >64 cores, starting 100 of threads in a container which is restricted to e.g. 2 CPUs leads to serious issues.

Until this behaviour is fixed in the JDK (there are plans to do so, but that's not a reality yet), I highly would recommend to allow for an absolute value to set (or use something like for maven-surefire with possible values like "1" (absoulte) or "1C" (multiplied by nr. of cores)).


rhuss:
You should disable the oraclejdk7 check for travis as this VM installation is broken there (hence the check failure)

mathieucarbou:
> Please remove the formatting to cleanup the diff and I'll merge. Thanks :-)

forget about it ;-) that was badly formatted

mathieucarbou:
Fixed through https://github.com/mycila/license-maven-plugin/pull/146

PR #144, opened 2019-01-31
Track the year of a file creation through license-maven-plugin-git

sic2:
Reason behind this PR:
- Sometimes you are required to include in your license header only the date of the file creation
- These changes allow us to look at the git history and find the year of creation for a given file

**Note**:
- Sometimes git is unable to figure out the date of creation of a file if this has moved/renamed without git being able to detect such change. 

A potential future solution could be to avoid dates to be overwritten. 
I would leave that for another PR though.


sic2:
- Reverted back the project version + the jgit version (which was not compliant with Java 7)
- Reverted back the imports order (my IntelliJ IDE setup was doing it for me)
- Reduced a bit of duplicated code further

PR #170, opened 2020-06-03
License Sets feature - allows one execution to check different licenses

adamretter:
This PR adds a new feature via config `<licenseSets>`.

You can now specify multiple `<licenseSet>` each with a different license, basedir, include and excludes paths, etc.

This means that a single execution of the plugin can check different files against different licenses. This is very useful in projects where you have a mix of source code licenses... perhaps because some files have been copied in from other compatible open source projects.

The changes are backwards compatible, so that the old config style also still works too.

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
@@ -44,16 +44,7 @@ import org.xml.sax.InputSource;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.ServiceLoader;
+import java.util.*;
 import java.util.concurrent.CompletionService;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutionException;
@@ -72,26 +63,35 @@ import static java.util.Arrays.deepToString;
  */
 public abstract class AbstractLicenseMojo extends AbstractMojo {
 
+    @Parameter
+    public LicenseSet[] licenseSets;
+
     /**
      * The base directory, in which to search for project files.
      */
-    @Parameter(property = "license.basedir", defaultValue = "${basedir}", required = true)
-    public File basedir;
+    @Parameter(property = "license.basedir", defaultValue = "${basedir}", alias = "basedir", required = true)
+    public File baseBasedir;
 
     /**
      * Location of the header. It can be a relative path, absolute path,
      * classpath resource, any URL. The plugin first check if the name specified
      * is a relative file, then an absolute file, then in the classpath. If not
      * found, it tries to construct a URL from the location.
+     *
+     * @deprecated use {@link LicenseSet#header}
      */
-    @Parameter(property = "license.header")
-    public String header;
+    @Deprecated
+    @Parameter(property = "license.header", alias = "header")
+    public String legacyConfigHeader;
 
     /**
      * Header, as text, directly in pom file. Using a CDATA section is strongly recommended.
+     *
+     * @deprecated use {@link LicenseSet#inlineHeader}
      */
-    @Parameter(property = "license.inlineHeader")
-    public String inlineHeader;
+    @Deprecated
+    @Parameter(property = "license.inlineHeader", alias="inlineHeader")
+    public String legacyConfigInlineHeader;
 
     /**
      * Specifies additional header files to use when checking for the presence
@@ -102,9 +102,12 @@ public abstract class AbstractLicenseMojo extends AbstractMojo {
      * <br>
      * When using remove goal, this property will be used to detect all valid
      * headers that also must be removed.
+     *
+     * @deprecated use {@link LicenseSet#validHeaders}
      */
-    @Parameter
-    public String[] validHeaders = new String[0];
+    @Deprecated
+    @Parameter(alias = "validHeaders")
+    public String[] legacyConfigValidHeaders = new String[0];
 
     /**
      * Alternative to `header`, `inlineHeader`, or `validHeaders`
@@ -114,23 +117,29 @@ public abstract class AbstractLicenseMojo extends AbstractMojo {
      * and have them concatenated together by the plugin. This
      * allows you to maintain each distinct license header in
      * its own file and combined them in different ways.
+     *
+     * @deprecated use {@link LicenseSet#multi}
      */
+    @Deprecated
     @Parameter
-    public Multi multi;
+    public Multi legacyConfigMulti;
 
     /**
      * Allows the use of external header definitions files. These files are
      * properties like files.
      */
-    @Parameter
-    public String[] headerDefinitions = new String[0];
+    @Parameter(alias = "headerDefinitions")
+    public String[] baseHeaderDefinitions = new String[0];
 
     /**
      * HeadSections define special regions of a header that allow for dynamic
      * substitution and validation
+     *
+     * @deprecated use {@link LicenseSet#headerSections}
      */
-    @Parameter
-    public HeaderSection[] headerSections = new HeaderSection[0];
+    @Deprecated
+    @Parameter(alias = "headerSections")
+    public HeaderSection[] legacyConfigHeaderSections = new HeaderSection[0];
 
     /**
      * You can set here some properties that you want to use when reading the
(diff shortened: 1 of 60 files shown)


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -72,26 +63,35 @@
...
+    @Parameter
+    public LicenseSet[] licenseSets;
+
     /**
      * The base directory, in which to search for project files.
      */
-    @Parameter(property = "license.basedir", defaultValue = "${basedir}", required = true)
-    public File basedir;
+    @Parameter(property = "license.basedir", defaultValue = "${basedir}", alias = "basedir", required = true)
+    public File baseBasedir;
```
mathieucarbou, line 73: why baseBasedir ? could we have a better name ?

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -127,38 +133,47 @@
      * replaced when the header file is read by those you specified in the
      * command line, in the POM and in system environment.
      */
-    @Parameter
-    public Map<String, String> properties = new HashMap<String, String>();
+    @Parameter(alias = "properties")
+    public Map<String, String> baseProperties = new HashMap<String, String>();
```
mathieucarbou, line 137: what "base" pefix means ?

license-maven-plugin/src/test/java/com/mycila/maven/plugin/license/LicenseSetTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.maven.plugin.testing.stubs.MavenProjectStub;
+import org.junit.Test;
+
+import java.io.File;
+
+public class LicenseSetTest {
+
+    @Test
+    public void multipleLicenseSets() throws Exception {
```
mathieucarbou, line 26: I would definitely add more tests, especially to test when the user mixes some "legacy" or "base" config with 1 and/or 2 sets.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
adamretter: So the `base` prefix is used to imply that this value is the base value, i.e. it can be overriden in `licenseSet`. If its not provided in a `licenseSet` then the licenseSet uses the base value ;-)

mathieucarbou: ok!
And what happens if some "licenseSet" definitions have intersections ? This is eagearly detected ? Or the formatter / checker will run on the same files with different licenses ?

adamretter: I am not quite sure what you mean by "intersections"?  ....The entire logic that was there previously, is now run once per-licenseSet.

mathieucarbou: If I understand correctly, `licenseSets` is equivalent to sequence of a plugin definition (headers + include/excludes).

So 2 questions:

1. What if the user defines different kind of licenses but turns out some set have the same inclusions ?
2. What's the different with defining multiple maven execution sets, each one with its own configuration and with different ids ?

```xml
<plugin>
    <groupId>com.mycila</groupId>
    <version>X.Y.ga</version>
    <configuration>
[global config here]
    </configuration>
    <executions>
        <execution>
            <id>first</id>
            <goals>
                <goal>check</goal>
            </goals>
            <configuration>
[config here for first run with its own includes]
            </configuration>
        </execution>
        <execution>
            <id>secondid>
            <goals>
                <goal>check</goal>
            </goals>
            <configuration>
[config here for second run with its own includes]
            </configuration>
        </execution>
    </executions>
</plugin>
```

adamretter: > What if the user defines different kind of licenses but turns out some set have the same inclusions ?

As set will operate on whatever inclusions it has. So if two sets share inclusions, both will be applied to the same included files.

> What's the different with defining multiple maven execution sets, each one with its own configuration and with different ids ?

Ahhhh... Glad you asked! So there is a longer story behind my motivation. Originally I was using multiple execution.... I added this PR because I was told by the Maven maintainers that using such executions in that manner was not a good idea, and that I should improve the plugin itself - see this PR to Maven for background - https://github.com/apache/maven/pull/357

mathieucarbou: How is it a good idea ? Maven is clearly allowing this: to have the same plugin executed several times on different execution ids and eventually phases.  What use case you cannot solve by using that plus eventually profiles ?

adamretter: > Maven is clearly allowing this

Yes I agree, it does allow that.

I don't really understand the reasoning of the response I got from the Maven project, but the advice there was that I should not use multiple executions for what I am doing, but should instead enhance the plugin to perform operations on sets of things.

Hopefully this feature is acceptable for you?

adamretter:
@mathieucarbou Thank you :-)
I have rebased, and improved the variable naming and javadoc. I will add some more tests shortly...

adamretter:
@mathieucarbou I have added some Integration Tests to see how you feel about them. If you are happy with them let me know, and I will add a few more for mixing licenseSets and legacy config etc...

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
mathieucarbou: Well, If we have the choice between solving the use case using what Maven provides and solving it by modifying and deprecating the plugin to mimic Maven's behavior, I prefer the former.
The point is not that the code is not acceptable, the point is that it solves a problens that shouldn't be solved as part of a plugin.
Tnis use case is mainly to be able to run multiple times an execution of a plugin. There is no Maven plugin that I know of that has this execution set mechanism coded in itself, especially because Maven is done to allow people to provided their own configuration inheritance with plugin configuration and different executions that can be configured too.
Where did you understand that the Maven team does not encourage to use multiple executions ? I didn't see such thing...  It is part of the inheritance mechanism of Maven also.

adamretter: > There is no Maven plugin that I know of that has this execution set mechanism coded in itsel

I just did a quick survey of one large Maven project here and it seems that there are actually lots of Maven plugins that allow you to do set based operations. I just found these:

1. maven-assembly-plugin takes a set of descriptors.
2. appassembler-maven-plugin takes a set of daemons.
3. copy-maven-plugin takes a set of resources.
4. maven-clean-plugin takes a set of filesets.
5. xml-maven-plugin takes a set of transformsets.
6. copy-rename-maven-plugin takes a set of filesets.
7. maven-dependency-plugin takes a set of artifactItem.

> Where did you understand that the Maven team does not encourage to use multiple executions

I got that impression from the feedback I was given here - https://github.com/apache/maven/pull/357

mathieucarbou: Yes you are right!
Some are just filesets, equivalent to what we have in MLP.  The closest would be the assembly plugin...
But you are right that from a usage perspective this is fare more friendly to provide that...
You convinced me ;-)

mathieucarbou:
@adamretter : please rebase, ensure everything is tested and let me know when I can merge :-)

adamretter:
@mathieucarbou Okay thanks :-)

I have now added all the integration tests. It might be best to first merge https://github.com/mycila/license-maven-plugin/pull/171 and then merge this one.

adamretter:
@mathieucarbou I have added some information to the README.md. Thanks.

mathieucarbou:
Merging! Thanks s lot for your hard work 👍

adamretter:
@mathieucarbou Amazing stuff. Thank you for getting these all merged.

Would you like to have a release? Would it be useful for me to help?

mathieucarbou:
> @mathieucarbou Amazing stuff. Thank you for getting these all merged.
> 
> Would you like to have a release? Would it be useful for me to help?

I will take care of it since it requires some creds for sonatye . I just want to wait for PR #172 to be merged first :-)

adamretter:
@mathieucarbou Again, thank you :-)

I do have Sonatype creds, but I think you would have to ask Joel to add me for the groupId.

mathieucarbou:
> @mathieucarbou Again, thank you :-)
> 
> I do have Sonatype creds, but I think you would have to ask Joel to add me for the groupId.

I didn't know it was possible!

PR #179, opened 2020-07-15
Bring in a number of default ignores / remove one that was duplicated

hazendaz:


hazendaz:
This list probably could grow forever.  I added most common ones I run into over 500+ repos at my day job.  I'm sure there are many more and this is just the tip of the iceburg.  Let me know what you think.

On the '.idea' one, I have been getting reports that 3.0 is licensing files in the folder so I suspect since this was only one not in the typical style as others it was likey some user error but see no harm in ensuring any .idea folder elsewhere is also ignored.

The xls entry was in the file already under docs.

mathieucarbou:
Nice! Thanks a lot!
I will merge as soon as I see a successfy build.

hazendaz:
@mathieucarbou Sorry I forgot to commas.  Fixed that.  I also added **/node/** to be skipped.  Same reason as node_modules.  Thanks.

After this is in, is it possible for you to run a RC2?  I have a number of people that want to try this out.  More so for the .idea change as it was confirmed to me that idea will stick the idea folder into any folder that contains node/node modules so it will show up many times.

Beyond that I think I have a few more things but need to run some wide testing before I suggest more.  Appreciate it!

hazendaz:
@mathieucarbou  Ping!

mathieucarbou:
sounds good! thanks a lot!

PR #186, opened 2020-09-02
[docType] Use slashstar for groovy, kotlin, and scala as proper usage

hazendaz:
These need to align with java usage.  All three were confirmed that they should be slash star.


hazendaz:
@mathieucarbou If you look at these three language types on github, each have their own source files for the language types using slash star now.

[kotlin example](https://github.com/JetBrains/kotlin/blob/master/core/descriptors.jvm/src/org/jetbrains/kotlin/resolve/jvm/inlineClassManglingRules.kt)

[groovy example](https://github.com/apache/groovy/blob/master/src/main/groovy/groovy/beans/ListenerList.groovy)

[scala example](https://github.com/scala/scala/blob/2.13.x/src/compiler/scala/reflect/macros/compiler/DefaultMacroCompiler.scala)

mathieucarbou (approved):

mathieucarbou:
this will break backward compatibility, but I think this is ok for version 4. This will be in release `4.0.rc3`, just before the official 4.0.

hazendaz:
@mathieucarbou Correct, same breaking change that was done with java so it matches.

PR #198, opened 2021-04-17
Jenkinsfile is source code

peterlynch:
Jenkinsfile is intended to be checked into SCM. As such it should contain a license just like any other source file.

https://www.jenkins.io/doc/book/pipeline/jenkinsfile/

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/Default.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/Default.java
@@ -198,9 +198,6 @@ public final class Default {
         "**/*.ods",
         "**/*.pdf",
 
-        // Jenkins
-        "**/Jenkinsfile",
-
         // Travis
         "**/.travis.yml",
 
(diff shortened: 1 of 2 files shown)


mathieucarbou (approved):

hazendaz:
@mathieucarbou Few questions here.  Testing in at least windows, Jenkinsfile will just throw a warning because it has no file extension.  What style would it have?  I presume those must be both figured out with standard code and not seeing how here.

peterlynch:
My bad. I meant to follow up with pr adding slash star style format for
jenkinsfile...

mathieucarbou:
@peterlynch could you please send a PR asap to rollback the support for Jenkins files ? I will be able to release a fix and it will give time for proper support and testing. I think the issue comes from the fact that license style detection is done based on the extension and this file has no extension.
I think that the code returning the extension of a file perhaps could return the file name if no extension is set so that we can correctly determine header styles based on either extension or fixed file names.

Thanks for the report @hazendaz 👍

peterlynch:
I can try to do this in the next few days

mathieucarbou:
@hazendaz : meanwhile, you could easily fix that by adding the Jenkins files in your list of excluded files I think.
Here is the commit FYI: https://github.com/mathieucarbou/license-maven-plugin/commit/33c7cd593503192283c30288154b745bbd47807f

hazendaz:
@mathieucarbou Yes, thanks.  I did exactly that.  Not a huge deal for now.  Will be nice to see ability to license such files as we have a number of them that don't fit nicely currently.  Appreciate it!

PR #214, opened 2021-08-20
Issue 213 - do not corrupt XML files

### LINKED ISSUE #213: XML files become corrupted if existing license header closes comment on the same line
Running `license:format` using `4.1` **corrupts XML files** if the existing (in XML file) license header looks like this:

```xml
<!-- All content copyright (c) 2003-2021 Bla, Inc., except as may
  otherwise be noted in a separate copyright notice. All rights reserved. -->
```

It works correctly if the license header looks like this: (close comment on a new line)
```xml
<!-- All content copyright (c) 2003-2021 Bla, Inc., except as may
  otherwise be noted in a separate copyright notice. All rights reserved. 
-->
```

With the first snippet, the resulting file looks like this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!--

    All content copyright (c) 2003-2021 Bla, Inc., except as may otherwise be
    noted in a separate copyright notice. All rights reserved.

-->
</ns:mytag>
```
(The last xml close tag is the only survivor)

Config:
```xml
 <plugin>
  <groupId>com.mycila</groupId>
  <version>4.1</version>
  <configuration>
    <inlineHeader><![CDATA[
All content copyright (c) 2003-2021 Bla, Inc., except as may otherwise be
noted in a separate copyright notice. All rights reserved.
        ]]></inlineHeader>
    <mapping>
      <java>SLASHSTAR_STYLE</java>
    </mapping>
    <excludes>
      <exclude>Jenkinsfile</exclude>
      <exclude>**/pom.xml</exclude>
    </excludes>
  </configuration>
</plugin>
```

FYI:
* Same behavior with `3.0`
* Problem does not occur with `XML_PER_LINE` mode

akomakom:
/closes #213

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderParser.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderParser.java
@@ -167,11 +167,11 @@ public final class HeaderParser {
             boolean foundEnd = false;
             do {
                 inPlaceHeader.append(line.toLowerCase());
-                line = fileContent.nextLine();
                 if (headerDefinition.isMultiLine() && headerDefinition.isLastHeaderLine(line)) {
                     foundEnd = true;
                     break;
                 }
+                line = fileContent.nextLine();
             }
             while (line != null && line.startsWith(before));
             // skip blank lines after header text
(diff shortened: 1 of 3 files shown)


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderParser.java
```diff
@@ -167,11 +167,11 @@ private boolean hasHeader() {
             boolean foundEnd = false;
             do {
                 inPlaceHeader.append(line.toLowerCase());
-                line = fileContent.nextLine();
```
akomakom, line 170: If `end tag` is on the current line, it'll be missed and the whole file will be treated as the header

mathieucarbou (approved):

mathieucarbou:
Thanks a lot!

PR #240, opened 2021-10-20
Support injecting Copyright Years based on file existence

rvesse:
For some Copyright policies the years in the header should reflect only
the existence of that file not the project as a whole.  While this can
be achieved by using `${license.git.copyrightCreationYear}` and
`${license.git.copyrightLastYear}` in the header template you end up
generating awkward headers for some files e.g. `2021-2021`.  And you
cannot use `${license.git.copyrightYears}` for this use case because that
will use the project inception year giving you something like `2019-2021`.

This commit adds a new calculated property
`${license.git.copyrightExistenceYears}` which provides this information
and may be used in header templates where this behaviour is desired.  This
should basically address the issue described in adobe/S3Mock#241 

It also fixes a bug in `GitLookup` whereby if you had modified the file
it would incorrectly calculate the year range (because the creation year
was always defaulted to current year for modified files) and falsely flag
a license header as missing (or reformat unncessarily).

### DIFF
--- a/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightRangeProvider.java
+++ b/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightRangeProvider.java
@@ -35,6 +35,7 @@ public class CopyrightRangeProvider extends GitPropertiesProvider implements Pro
 
   public static final String COPYRIGHT_LAST_YEAR_KEY = "license.git.copyrightLastYear";
   public static final String COPYRIGHT_CREATION_YEAR_KEY = "license.git.copyrightCreationYear";
+  public static final String COPYRIGHT_EXISTENCE_YEARS_KEY = "license.git.copyrightExistenceYears";
   public static final String COPYRIGHT_YEARS_KEY = "license.git.copyrightYears";
   public static final String INCEPTION_YEAR_KEY = "project.inceptionYear";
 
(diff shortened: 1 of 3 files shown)


mathieucarbou:
Sorry for the delay, I will merge and issue a 4.2.rc2.
Thanks!

mathieucarbou (approved):

rvesse:
@mathieucarbou No worries, thanks for accepting this PR 👍

mathieucarbou:
@rvesse : FYI a (minor) bug was discovered in a part implemented by @rremer so I will probably wait a little more before a new rc to get the chance to fix that. But if you need that sooner please let me know...

rvesse:
@mathieucarbou No great rush on my end

dbwiddis:
I just discovered this new feature and I'd like to implement it.  Any chance of rc2 soon-ish?  I tried using SNAPSHOT but it appears only the metadata is there and no actual release files.

mathieucarbou:
> I just discovered this new feature and I'd like to implement it.  Any chance of rc2 soon-ish?  I tried using SNAPSHOT but it appears only the metadata is there and no actual release files.

Yes it's planned next week 😊

dbwiddis:
I'm running into an odd issue with this plugin.  It's rewriting all headers and removing the earlier year, but only on OpenBSD.

For a sample file:
```
mvn verify
```
Changes the file thus:
```
- * Copyright (c) 2020-2021 ...
+ * Copyright (c) 2021 ...
```
This of course modifies the file so running it again turns it all into 2021-2022 on the next pass...

Should I create a new issue?  Anything else I can help test/debug?  Some configuration I'm missing?

dbwiddis:
Ah, figured it out.  On that VM I had a shallow clone so the year determination was only going back a few hundred commits.

So, the follow on question: would it be possible for the plugin to detect the presence of a shallow clone and not rewrite if that is the case?

mathieucarbou:
@dbwiddis : I think this is a legitimate use case, yes. We could add an option in the plugin. I have created #252.
I won't / cannot work on it now, if you would like to contribute it, you are welcomed to 👍 
The new RC was done on Monday, by the way ;-)

Issue #242, opened 2021-11-19
Header is added instead of being replaced, if formats differ


cns-solutions-admin:
If in multi line patterns the before pattern changes, the header is added instead of being replaced.

If the custom header style looks like this:

/************
**
************/

and the current header is

/**
 *
 */

then only the first line is replaced with the new header.

cns-solutions-admin:
Fixed in PR #243 by introducing a new optional regex.

mathieucarbou:
Ok I understand a bit more... Let me review...

AdrienHorgnies:
Hi,

I have the same problem even when the formatting is exactly the same.
In the code block below, you can see that the new header has been prepended right above the previous one :

```java
/*
 * SWIFT (c) 2022. All rights reserved.
 *
 * This software and its associated documentation contain
 * proprietary, confidential and trade secret information of
 * S.W.I.F.T. SCRL, and except as provided in your contractual
 * arrangements with S.W.I.F.T. SCRL.
 * a) no part may be reproduced, made available,
 *    adapted or translated in any form or by any means, and
 * b) this software may only be installed and used by a duly
 *    registered SWIFT User, SWIFT Partner or Service Bureau
 *    in compliance with the terms of the licence with SWIFT.
 *    This software may only be used to access and use the
 *    SWIFT services and products.
 * For more information about what you may or may not do with
 * this software and any related documentation, including any
 * limitations on warranties and/or remedies, refer to your
 * relevant contractual arrangements with SWIFT.
 */
/*
 * SWIFT (c) 2021. All rights reserved.
 *
 * This software and its associated documentation contain
 * proprietary, confidential and trade secret information of
 * S.W.I.F.T. SCRL, and except as provided in your contractual
 * arrangements with S.W.I.F.T. SCRL.
 * a) no part may be reproduced, made available,
 *    adapted or translated in any form or by any means, and
 * b) this software may only be installed and used by a duly
 *    registered SWIFT User, SWIFT Partner or Service Bureau
 *    in compliance with the terms of the licence with SWIFT.
 *    This software may only be used to access and use the
 *    SWIFT services and products.
 * For more information about what you may or may not do with
 * this software and any related documentation, including any
 * limitations on warranties and/or remedies, refer to your
 * relevant contractual arrangements with SWIFT.
 */
```

And I have the below configuration :

```xml
<plugin>
    <groupId>com.mycila</groupId>
    <artifactId>license-maven-plugin</artifactId>
[...]

PR #243, opened 2021-11-19
Add regex for line comparison

cns-solutions-admin:
This solves #242 by adding a new optional regex to compare the lines other than the first for multiline headers.


cns-solutions-admin:
Also converts a inline header style's name to lower case, otherwise the comparison (and the example in the docs) fails.

mathieucarbou:
Hello,
Why changing the header definition didn't work for your use case ?

cns-solutions-admin:
Changing the header definition did not work, because the detection of header lines was orginally done by comparing lines with startsWith(beforeEachLine.rtrim())
the old " * something" does not start with "**".

By introducing a (optional) RegEx for the header lines, this and most other use cases can be covered.
It also fits with the current configuration, where we already have RegExs for start and end line.

docs/index.md
```diff
@@ -538,14 +538,14 @@ In license-maven-plugin, each header style is defined by patterns to detect it a
...
-    <endLine> */</endLine>
+    <beforeEachLine xml:space="preserve"> * </beforeEachLine>
+    <endLine xml:space="preserve"> */</endLine>
     <afterEachLine></afterEachLine>
     <!--skipLine></skipLine-->
     <firstLineDetectionPattern>(\s|\t)*/\*.*$</firstLineDetectionPattern>
     <lastLineDetectionPattern>.*\*/(\s|\t)*$</lastLineDetectionPattern>
     <allowBlankLines>false</allowBlankLines>
-    <isMultiline>true</isMultiline>
+    <multiline>true</multiline>
```
mathieucarbou, line 548: Please remove any change regarding multiline: I've correctly fixed in https://github.com/mathieucarbou/license-maven-plugin/pull/247

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderDefinition.java
```diff
@@ -34,6 +36,7 @@
 
   private Pattern skipLinePattern;
   private Pattern firstLineDetectionPattern;
+  private Pattern otherLineDetectionPattern;
```
mathieucarbou, line 39: I would prefer that the name is called `middleLineDetectionPattern`

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderDefinition.java
```diff
@@ -209,6 +257,7 @@ public void validate() {
     check("endLine", this.endLine);
     check("afterEachLine", this.afterEachLine);
     check("firstLineDetectionPattern", this.firstLineDetectionPattern);
+    // other line detection pattern can be null
```
mathieucarbou, line 260: no need for this comment: it's like `beforeEachLine` ;-)

license-maven-plugin/src/test/java/com/mycila/maven/plugin/license/header/HeaderParserTest.java
```diff
@@ -101,4 +101,33 @@ public void test_parsing_xml6() throws Exception {
...
+     HeaderParser parser = getCustomJavadocHeaderParser("src/test/resources/doc/java3.txt");
+     assertTrue(parser.gotAnyHeader());
+     assertEquals("package", parser.getFileContent().getContent().substring(parser.getEndPosition()).trim().substring(0, 7));
+  }
+  
+  private HeaderParser getCustomJavadocHeaderParser(final String fileName) {
+     final HeaderDefinition headerDefinition = new HeaderDefinition(
+           "javadoc2_style", "/**", " ** ", " **/", "", null, "^\\s*/\\*.*$", "^\\s*\\*.*$", "^.*\\*/\\s*$", false, true, false);
+     return new HeaderParser(new FileContent(new File(fileName), System.getProperty("file.encoding")), headerDefinition, new String[]{"copyright"});
+  }
```
mathieucarbou, line 130: You would need also to have a more complete test (like `CompleteMojoTest` does) to also test the other plugin actions, like really testing the issue you want to fix.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderDefinition.java
```diff
@@ -61,13 +64,40 @@ public HeaderDefinition(String type,
...
+                          String firstLineDetectionPattern, String otherLineDetectionPattern, String lastLineDetectionPattern,
+                          boolean allowBlankLines, boolean isMultiline, boolean padLines) {
     this(type);
     this.firstLine = firstLine;
     this.beforeEachLine = beforeEachLine;
     this.endLine = endLine;
     this.afterEachLine = afterEachLine;
     this.skipLinePattern = compile(skipLinePattern);
     this.firstLineDetectionPattern = compile(firstLineDetectionPattern);
+    this.otherLineDetectionPattern = compile(otherLineDetectionPattern);
```
mathieucarbou, line 100: you would need to update `setPropertyFromString(...)` too for xml header definition.

mathieucarbou (changes requested):
Please also squash and rebase. Thanks 👍

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderDefinition.java
cns-solutions-admin: Except it is for all lines except the first line - the last line must also pass this pattern (or the beforeEachLine test).
Finding the best name is often the most difficult task ;-)

mathieucarbou: > the last line must also pass this pattern

Why ? There is a pattern for the last line.

cns-solutions-admin: But the code still checks beforeEachLine. I didn't want to change the logic and the RegEx happens when the startsWith(beforeEachLine) check happens.

PR #247, opened 2021-11-26
Correctly fix the multiLine tag

mathieucarbou:


mathieucarbou:
@cns-solutions-admin FYI: this PR correctly change the `multiLIne` tag and supports backward compat'

PR #253, opened 2022-01-15
Optionally warn (once) if repository is shallow/sparse

### LINKED ISSUE #252: Detect shallow clones for git plugin to avoid reading a wrong date range (which first year being wrong)
`git rev-parse --is-shallow-repository` which will print false (not shallow) or true (shallow):

See [#240](https://github.com/mathieucarbou/license-maven-plugin/pull/240#issuecomment-1010734242) for a use case and problem.

dbwiddis:
Fixes #252.

Adds a new parameter `skipIfShallow` that will have the same effect as `skip` if executed on a shallow clone.


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -295,6 +296,15 @@
...
   public boolean skip = false;
 
+  /**
+   * Whether to skip the plugin execution on a shallow git clone.
+   * <p>
+   * We recommend setting this to {@code true} if you are generating copyright
+   * ranges from file creation dates in git history.
+   */
+  @Parameter(property = "license.skipIfShallow", defaultValue = "false")
+  public boolean skipIfShallow = false;
```
hazendaz, line 306: @dbwiddis Remove the = false here as that is java default and when defined is initializing twice.

dbwiddis: OK but I copied verbatim from `skip` 8 lines above, and `quiet` 8 lines below is the same format. Should I fix those too?

hazendaz: I'm going to be fixing all those issues shortly :)

hazendaz:
@dbwiddis I'm adding github actions ci in...waiting on abilty to self merge as I'm otherwise waiting.  I trust that a bit more than what all this seems to be trying to pull off in travis CI so I'll come back to this afterwards.  Hopefully later this weekend.  As to the change, I feel like the default may need flipped but false to start with while its being floated around seems like a logic first step.  We did something similar on formatter maven plugin and even flipped the flag for our next release after it was proven out.  Anyway, this is useful in regards to your other findings ;)

dbwiddis:
> I feel like the default may need flipped 

I went back and forth on a few other ideas.  The shallow history is only really relevant if you're generating copyright range using `${license.git.copyrightCreationYear}` or the new `${license.git.copyrightExistenceYears}`.  I pondered checking for those definitions in the template and issuing a log warning but figured it would get lost in the spam of header rewrites.

Potentially this setting should be mentioned in the documentation for those template substitutions, though.

dbwiddis:
Converting this to a draft because I think this can be done one of two ways without the `Runtime.exec()` call:
1. It does look like JGIt has functionality to detect a shallow clone (PackStatistics depth > 0)
2. We can also test for the existence of `.git/shallow` in the filesystem

mathieucarbou:
Thanks guys for looking into that 👍

dbwiddis:
OK, new approach!  I think this makes a lot more sense, causing a failure when it should...

 - The parameter is now `failIfShallow` and defaults to `true`.
 - Detects shallow clone by testing for existence of `.git/shallow` file.
 - If `failIfShallow` is `true` causes an exception to be thrown only when attempting to fetch properties requiring full git history
   - original author/email
   - copyright dates that are not the inception year

dbwiddis:
Hmm, obviously I messed something up. Will hunt down that NPE.

dbwiddis:
It's throwing NPE on this line (two places):
```
if (mojo.failIfShallow && gitLookup.isShallowRepository()) 
```
After my latest fix it is impossible for `gitLookup` to be null.  Is `mojo` null in this case?  I'm not sure how I can access the value of the `failIfShallow` property.

dbwiddis:
Ah, I see the problem, the test is passing `null` for the mojo:
```
Map<String, String> actual = provider.getAdditionalProperties(null, props, document);
```
I'll instantiate a new somethingMojo in the test...

dbwiddis:
Switched back to draft for 2 discussion points:
1. At present we'd throw an exception for shallow repo for the `project.inceptionYear`.  That doesn't require git history so I'd like to refactor that substitution out to its own provider.
2. Should I null-check `repo` in the provider and assume default, e.g., `true` setting in that case? Is there any real world use case where null is a valid argument?

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightAuthorProvider.java
```diff
@@ -56,12 +56,15 @@ public CopyrightAuthorProvider() {
     try {
       Map<String, String> result = new HashMap<String, String>(3);
       GitLookup gitLookup = getGitLookup(document.getFile(), properties);
-
+      
+      if (mojo.failIfShallow && gitLookup.isShallowRepository()) {
+        throw new RuntimeException("Shallow repository detected.");
+      }
```
mathieucarbou, line 62: After some thinking, I am wondering if this is the right way of implementing that...

We should keep backward compatibility. So for users having a “working” or “acceptable” setup with shallow clones, it should continue to work as-is. The current situation can be accepted in many cases where the "badly" computed values would turn out matching the ones of the first commit.

For the remaining **new** users, I would probably avoid throwing and instead returning a Map, either empty or without the properties that cannot be correctly computed, and if a property cannot be computed, then log a message as info, if the user wants.

This would be a combination of 2 settings (we can find better names of course):

- `mojo.shallowSupport` => default to false, Activate or not correct support for shallow clones
- `mojo.warnUnresolvableProperties` => default to false, If `mojo.shallowSupport` is set to true and this property set to true, then we would log a warning for all the properties that are skipped because they cannot be computed
- `mojo.failOnUnresolvableProperties` => default to false, If `mojo.shallowSupport` is set to true and this property set to true, then we would fail the build if a property cannot be computed because of a shallow clone

I'm not sure a lot of people would activate 3. Only 1 and 2 would be enough. If someone is using a property that cannot be computed, it will be seen as the license check will fail the build.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -76,7 +77,8 @@ public GitLookup(File anyFile, DateSource dateSource, TimeZone timeZone, int che
     super();
     this.repository = new FileRepositoryBuilder().findGitDir(anyFile).build();
     /* A workaround for  https://bugs.eclipse.org/bugs/show_bug.cgi?id=457961 */
-    this.repository.getObjectDatabase().newReader().getShallowCommits();
+    // Also contains contents of .git/shallow and can detect shallow repo
+    this.shallow = !this.repository.getObjectDatabase().newReader().getShallowCommits().isEmpty();
```
mathieucarbou, line 81: This reader was not closed here... We missed a try/resource I think ?

mathieucarbou:
> 2\. Should I null-check `repo` in the provider and assume default, e.g., `true` setting in that case? Is there any real world use case where null is a valid argument?

I I maybe lacking context (I don't understand fully the question), but if it is regarding injected Maven properties, then we have to stub them in our test. In real life they won't be null.

Or the question is regarding the existence or not of a git repo when this plugin is used ?

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
dbwiddis: Yeah, I saw the warning in my IDE, but everywhere else I see this workaround posted, nobody is using try with resources and it's not obvious to me that it needs to be closed, and possibly closing it would negate whatever the workaround is.  Given my earlier mistakes on resources I decided to leave it as-is (other than collecting some side-effect information from it).

dbwiddis:
> I maybe lacking context (I don't understand fully the question), but if it is regarding injected Maven properties, then we have to stub them in our test. In real life they won't be null.

That was essentially the question. The context was that in the test cases (here) we were passing null which wasn't a problem when we never referred to the mojo, but is now.  So my "test fix" creating a new mojo was the appropriate way to address this.

I'll mark this as ready again, pending further discussion!

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightAuthorProvider.java
dbwiddis: > We should keep backward compatibility. So for users having a “working” or “acceptable” setup with shallow clones, it should continue to work as-is.

It still will, if that property is set false; users can quickly "disable" this if desired.  This does go back to @hazendaz comment [here](https://github.com/mathieucarbou/license-maven-plugin/issues/253#issuecomment-1013783799_) about flipped defaults, although the context of that was a warning rather than an exception.

I do like the gist of your suggestion, offering 3 possible outcomes based on setting:
 - Completely ignore with no warning
 - Warn only
 - Fail build

Personally I'd like to add a fourth option which I think would be ideal if we can figure out how:
 - Warn about possible date inaccuracy only on failed header check (default true); but if license is not missing/rewritten, remain silent

The reason I like this:
 - It fits my (and probably a very common) use case: develop on a machine with full git history, but execute tests on a machine with shallow history.  I include the `format` goal in my local builds so all the headers are supposed to be rewritten when I commit; there's no reason CI should bother rewriting them again.
 - it assists in debugging current behavior. In my initial testing, I ended up with tons of header rewrites only on one machine, and having a log warning "I'm rewriting these because I detected a changed date but the repo is shallow" would have saved me some debugging time.   

With my current solution, I could quickly replace the failure with a warning but it would warn on all files, probably too much logspam and would be "off" by default.  But a warning associated with every header rewrite I think is reasonable to default "on".

This leads to your "map" suggestion:
> a Map, either empty or without the properties that cannot be correctly computed

Except I think the best purpose/name of the map would not say "cannot be computed" but "not certain it was computed correctly".  This would fit my use case above (develop full depth, CI test shallow) where it is likely that the headers are already correct, so the ideal outcome in that case would be:
 - If an "uncertain" property exists and the header is present, assume it's accurate, don't try to change it if the repo is shallow, and silently ignore
 - If an "uncertain" property exists and the header is missing, add it, and warn about possible year in accuracy (by default)

Or, one other solution, going back to my initial submission, "warnOnShallow".  I think it's **entirely reasonable to have a (default true) single log warning that a repository is shallow**.  It can be generated in the `GetLookup` instantiation so it only occurs once (basically warning instead of generating the `shallow` field in the current iteration), and only if the git properties are used.  
 - This still helps debugging (in the case of mass header check warnings/rewrites) but doesn't otherwise change the existing behavior.
 - Users have existing workarounds (adding `-d"license.skip"` or using profiles in CI) to respond to this warning; and/or toggle it off.

> This general behavior then could be applied to the SVN plugin

And I'll figure out how to detect a "sparse" SVN history and do something similar there.

mathieucarbou: Whatever the solution : it **has** to be backward compatible and not fail existing setups or even log a lot of warnings in case of a project having a lot of modules.

The idea of having multiple properties matches what's done in other plugins too: warnIf..., failIf...

dbwiddis: > Whatever the solution : it **has** to be backward compatible and not fail existing setups or even log a lot of warnings in case of a project having a lot of modules.

OK, I think I'm pretty happy with the current iteration of it.  
1. It's backwards compatible, not failing existing setups
2. It doesn't log a lot of warnings. It only logs _one_.
  - And it only logs in the case that it's relevant, if the author info or creation/last modify year is requested.
  - It's therefore informational and useful for debugging 
3. Users can turn off that (one) warning line if they choose and "know what they're doing" with limited history.

mathieucarbou:
@dbwiddis I'm sorry I didn't notice all your last activity... I was wondering if your PR is ready to be reviewed again, and also if you could rebase it if it is ? Thanks a lot!

mathieucarbou (approved):

dbwiddis:
Rebased to master, although I don't think that's necessary when GitHub says "This branch has no conflicts with the base branch."

I think this PR is GTG but please note #300 which is essentially the same problem as I'm trying to solve here, but far less detectable, and I believe is a blocking bug to your next release. I'd suggest at least thinking through the solution to #300 before merging this to see if there are any changes needed.

Also note that I had submitted #286 for comment, which would be a workaround to this. I withdrew it because I think there's room for a much more flexible way of tokenizing the whole copyright line (basically everything after the word "copyright" up to the next newline should be allowed to be kept as is) which would also be a workaround for this.

mathieucarbou:
Thanks! Will merge.

PR #258, opened 2022-01-16
[ci] Drop travis CI

hazendaz:
rate limited, no longer open source friendly, and github actions is light years ahead.

Basic github actions added that does all that is necessary.  I'm not 100% clear why we are testing the release deployment features here but all that can be done via actions if we need something similar and probably only a few minutes worth of work.  Dropping it as travis is just bad at this point.


dbwiddis:
+1. Only good reason to still use Travis is if you need to test on ARM hardware, which is not relevant for this plugin.

PR #281, opened 2022-01-17
[github] Only run snapshot-report from mathieucarbou

hazendaz:
Preventing the site update from running on forks.


.github/workflows/snapshot-report.yml
```diff
@@ -9,6 +9,7 @@ on:
       - 'docs/**'
 jobs:
   generate-site:
+    if: github.repository_owner == 'mathieucarbou'
```
mathieucarbou, line 12: 👍

mathieucarbou (approved):

PR #283, opened 2022-01-17
Suppress resource leaks warnings in FileUtils

dbwiddis:
~`FileChannel` is not `Closeable`,~ (actually it is) ~so including it in a try-with-resources block does not close the underling streams.  Refactored to ensure the closeable object is referenced in try.~

Also implemented a try-with-resources block for the `BufferedReader` closure.

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
@@ -48,8 +48,8 @@ public final class FileUtils {
   }
 
   public static void write(File file, String content, String encoding) throws IOException {
-    try (FileChannel channel = new FileOutputStream(file).getChannel()) {
-      channel.write(ByteBuffer.wrap(content.getBytes(encoding)));
+    try (FileOutputStream outStream = new FileOutputStream(file)) {
+      outStream.getChannel().write(ByteBuffer.wrap(content.getBytes(encoding)));
     }
   }
 
@@ -68,18 +68,16 @@ public final class FileUtils {
   public static String[] read(final URL[] locations, final String encoding) throws IOException {
     final String[] results = new String[locations.length];
     for (int i = 0; i < locations.length; i++) {
-      final Reader reader = new BufferedReader(new InputStreamReader(locations[i].openStream(), encoding));
-      try {
+      try (Reader reader = new BufferedReader(new InputStreamReader(locations[i].openStream(), encoding))) {
         results[i] = IOUtil.toString(reader);
-      } finally {
-        reader.close();
       }
     }
     return results;
   }
 
   public static String read(File file, String encoding) throws IOException {
-    try (FileChannel in = new FileInputStream(file).getChannel()) {
+    try (FileInputStream inStream = new FileInputStream(file)) {
+      FileChannel in = inStream.getChannel();
       ByteArrayOutputStream baos = new ByteArrayOutputStream();
       in.transferTo(0, in.size(), Channels.newChannel(baos));
       return baos.toString(encoding);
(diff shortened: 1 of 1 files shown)


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
```diff
@@ -68,18 +68,16 @@ public static String read(URL location, String encoding) throws IOException {
   public static String[] read(final URL[] locations, final String encoding) throws IOException {
     final String[] results = new String[locations.length];
     for (int i = 0; i < locations.length; i++) {
-      final Reader reader = new BufferedReader(new InputStreamReader(locations[i].openStream(), encoding));
-      try {
+      try (Reader reader = new BufferedReader(new InputStreamReader(locations[i].openStream(), encoding))) {
```
mathieucarbou, line 71: 👍

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
```diff
@@ -48,8 +48,8 @@ private FileUtils() {
   }
 
   public static void write(File file, String content, String encoding) throws IOException {
-    try (FileChannel channel = new FileOutputStream(file).getChannel()) {
-      channel.write(ByteBuffer.wrap(content.getBytes(encoding)));
+    try (FileOutputStream outStream = new FileOutputStream(file)) {
+      outStream.getChannel().write(ByteBuffer.wrap(content.getBytes(encoding)));
```
mathieucarbou, line 52: Please revert: there was no leak here: the construction is right: we can convert a `FileOutputStream` to a `FileChannel` and only use and close the latter.

`FileChannel` implements `Channel`, which is `Closeable`. Besides, the javadoc clearly states that the channel will be closed, and as you can see in the `FileChannel` impl, its file descriptor is released and closed.

Regarding the usage of `FileOutputStream` and `FileChannel`, closing `FileOutputStream` will also close the channel, as stated in the javadoc of its `close()` method, and closing the `FileChannel` will also close the underlying file descriptor. Closing one or the other is fine.

This construction `new FileOutputStream(file).getChannel()` also prevents anyone from using the `FileOutputStream` and allows it to be GC-ed as soon as possible.

Side note: if there were such a leak in the plugin, that is running on very big projects, I think that it would have been discovered years go ;-)

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
```diff
@@ -68,18 +68,16 @@ public static String read(URL location, String encoding) throws IOException {
...
-        reader.close();
       }
     }
     return results;
   }
 
   public static String read(File file, String encoding) throws IOException {
-    try (FileChannel in = new FileInputStream(file).getChannel()) {
+    try (FileInputStream inStream = new FileInputStream(file)) {
+      FileChannel in = inStream.getChannel();
```
mathieucarbou, line 80: Same comment. Can be reverted.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
```diff
@@ -107,8 +105,10 @@ public static String remove(String str, String... chars) {
 
   public static void copyFileToFolder(File file, File folder) throws IOException {
     File dest = new File(folder, file.getName());
-    try (FileChannel inChannel = new FileInputStream(file).getChannel();
-         FileChannel outChannel = new FileOutputStream(dest).getChannel()) {
+    try (FileInputStream inStream = new FileInputStream(file);
+         FileOutputStream outStream = new FileOutputStream(dest)) {
+      FileChannel inChannel = inStream.getChannel();
+      FileChannel outChannel = outStream.getChannel();
```
mathieucarbou, line 111: Same comment. Can be reverted.

mathieucarbou (changes requested):

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
dbwiddis: Indeed you're right. I'd looked for `close()` and didn't see it but didn't look at the inheritance.

> Side note: if there were such a leak in the plugin, that is running on very big projects, I think that it would have been discovered years go ;-)

Yeah, probably true!  Got too eager trying to get rid of yellow triangles in my IDE. :-)

dbwiddis: Reverted and added some suppressed warnings to help the future me.

mathieucarbou (approved):

PR #310, opened 2022-03-24
Default to full depth for GitLookup

### LINKED ISSUE #151: Wrong `license.git.copyrightCreationYear` when the file has more than 10 commits
Although it is not yet released, I gave a try to version `3.1` mainly for its GIT _creation date_ feature.

I noticed the `RevWalk` is configured by default to consider only the last `10` commits. This is causing an issue when computing the creation date of a file that has more than 10 commits in its history. Consequence is the returned date is not the date of the initial commit as expected. Moreover, that date keeps changing as new changes are committed. This case is not catched by the test cases probably because the test repo does not have a file with more than 10 commits in its history.

### LINKED ISSUE #300: File creation year is inaccurate by default
I've been running rc2 for a while, using the copyright existence years implemented by #240, and it has been showing somewhat odd behavior, changing license headers on unrelated commits.  The behavior was similar to what I saw when the repository was shallow (addressed by #252) but I confirmed full depth.   But it's still showing wrong date ranges.

I finally dug into the code and saw that the [default commits count (for lookback) was set to 10](https://github.com/mathieucarbou/license-maven-plugin/blob/037cf8a911e0cbb2585bc533c8ed2db469af3197/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java#L51).   This is problematic for calculating creation year, which [simply walks from the oldest commit](https://github.com/mathieucarbou/license-maven-plugin/blob/f88cd50db2367704904aa68fd52d347828766637/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java#L145-L152) until it runs out of commits to check.  But "oldest" is set by the commit history depth (similar to shallow checkouts) where it [filters whole tree by commit number and then sorts it](https://github.com/mathieucarbou/license-maven-plugin/blob/f88cd50db2367704904aa68fd52d347828766637/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java#L204-L207).  

As noted in https://github.com/adobe/S3Mock/issues/241#issuecomment-674772745 it is possible to add a property for larger histories:
```
<license.git.copyrightLastYearMaxCommitsLookup>100000</license.git.copyrightLastYearMaxCommitsLookup>
```
Some issues with this:
* It's poorly named, implying it applies only to the "Last Year" while in fact it applies to the creation year
* There's no documentation that creation year is broken (in the same way a shallow clone breaks it) with the default
* I can't see where in my pom to put that; it doesn't seem to work in `<properties>` or `<configuration>` for the license plugin.  It does work for me on the command line with `-Dlicense.git.copyrightLastYearMaxCommitsLookup=100000`.

dbwiddis:
Fixes #300 using Option 1.  See the issue for justification and other possible solutions.

Also fixes #151 (closed as stale)

### DIFF
--- a/license-maven-plugin-git/README.md
+++ b/license-maven-plugin-git/README.md
@@ -20,7 +20,7 @@ Which properties is license-maven-plugin-git adding?
 
 The full git history for a file is required for accurate determination of the first commit (for the creation author name/email, creation year, or existence years). The plugin will warn if it detects a shallow git repository. If you are certain your shallow depth will still permit determination of these values you may suppress the warning by setting the parameter `license.warnIfShallow` to false.
 
-If you are not using the properties based on the first commit, you can slightly improve performance by limiting the git history for each file using the property `license.git.copyrightLastYearMaxCommitsLookup`.
+If you are using properties requiring the year of the last commit of a file but not using the "creation" properties requiring the first commit, you can improve performance by limiting the git history for each file using the property `license.git.maxCommitsLookup`. A value of 1 would provide the best performance but could be impacted if the last commit year is earlier than the year of previous commits.
 
 How to use license-maven-plugin-git
 -----------------------------------
--- a/license-maven-plugin-git/README.md
+++ b/license-maven-plugin-git/README.md
@@ -6,14 +6,21 @@ It is quite common that legal departments require updating of copyright years in
 What is license-maven-plugin-git?
 ---------------------------------
 
-* `license-maven-plugin-git` can optionaly be used with `license-maven-plugin` to bring in properties (esp. the year of the last change) from git.
+* `license-maven-plugin-git` can optionally be used with `license-maven-plugin` to bring in properties (esp. the year of the last change) from git.
 * The properties added by `license-maven-plugin-git` can be used in file header templates used by `license-maven-plugin`
 
 Which properties is license-maven-plugin-git adding?
 ----------------------------------------------------
 
 * `license.git.copyrightLastYear` - the year of the last change of the present file as seen in git history
-* `license.git.copyrightYears` - the combination of `project.inceptionYear` and `license.git.copyrightLastYear` delimited by a dash (`-`), or just `project.inceptionYear` if `project.inceptionYear` is eqal to `license.git.copyrightLastYear`
+* `license.git.copyrightYears` - the combination of `project.inceptionYear` and `license.git.copyrightLastYear` delimited by a dash (`-`), or just `project.inceptionYear` if `project.inceptionYear` is equal to `license.git.copyrightLastYear`
+* `license.git.CreationAuthorName` and `license.git.CreationAuthorEmail` - the name and email of the author of the first commit of the present file as seen in git history.
+* `license.git.copyrightCreationYear` - the year of the first commit of the present file as seen in git history
+* `license.git.copyrightExistenceYears` - Similar to `license.git.copyrightYears` but using `license.git.copyrightCreationYear` for the first year
+
+The full git history for a file is required for accurate determination of the first commit (for the creation author name/email, creation year, or existence years). The plugin will warn if it detects a shallow git repository. If you are certain your shallow depth will still permit determination of these values you may suppress the warning by setting the parameter `license.warnIfShallow` to false.
+
+If you are not using the properties based on the first commit, you can slightly improve performance by limiting the git history for each file using the property `license.git.copyrightLastYearMaxCommitsLookup`.
 
 How to use license-maven-plugin-git
 -----------------------------------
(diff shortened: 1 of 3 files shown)


mathieucarbou:
> This makes better sense but I suspect this will be a good performance hit and needs documented as the default. I'm ok with this but letting @mathieucarbou comment further.

I agree with the change to be the default, it makes complete sense.
The lookup would only be local anyway.
But I would make this configurable through a plugin setting + documented somewhere so that people impacted by the lost perf could set their own "max history count"

dbwiddis:
> This makes better sense but I suspect this will be a good performance hit 

Actually, it's not.  This value is a per-file history, so if your project has a large number of files with short history, they won't be impacted.   The existing default of 10 means that only files with more than 10 commits will take longer; and these are exactly the ones which would otherwise have an incorrect date.

This isn't a benchmark, but I just ran `'mvn license:check -D"license.git.copyrightLastYearMaxCommitsLookup=10"` (which replicates the default) on OSHI (2326 commits total) vs. other values from 1 to 150.  It's always pretty snappy on my *nix github action servers and locally on my mac, but these runs are on my MINGW64 terminal on windows which is notoriously slow.  The time for the default (10) depth was 2m14s.  The time for max history (125 or anything greater) was 2m32s.
```
Depth	Time (seconds)
1	70
2	86
3	93
4	102
5	111
6	117
7	112
8	120
9	124
10	134 <--- this is the current default
15	131
20	135
25	128
30	137
50	128
75	139
100	162
125	152
150	138
```
image

Also as another data point posted by @afranken in https://github.com/adobe/S3Mock/issues/241#issuecomment-674772745_

> adding `<license.git.copyrightLastYearMaxCommitsLookup>100000</license.git.copyrightLastYearMaxCommitsLookup>` to the configuration to enable finding a file's creation date even in large GIT histories and changing the line in the copyright template to ` Copyright ${license.git.copyrightCreationYear} Adobe.` works with version `4.0.rc2.

So I think the performance hit is negligible, and is only slower because it's actually finding the correct answer. 

> But I would make this configurable through a plugin setting + documented somewhere so that people impacted by the lost perf could set their own "max history count"

They already can set their own using the setting posted above, and that's exactly what is needed to workaround the bug.

I agree, though, that it should be documented, and I complained in #300 that it's not. Should it be a parameter? Should it be renamed to something shorter?

dbwiddis:
Another data point, demonstrating minimal performance impact:

My CI GHA (Ubuntu JDK17) uses that "essentially infinite" workaround: https://github.com/oshi/oshi/runs/5661865433?check_suite_focus=true

License check took 1m17s.   

Compare to the previous (default 10) setting: https://github.com/oshi/oshi/runs/5401533579?check_suite_focus=true

License check took 51s (and failed due to a handful of files losing their earliest year commit history).

There's really no good reason to limit the per-file commit history by default.   Consider the absurdity of this edge case (which I encountered something similar to prompting me to look into this):  a file created 10 years ago in 2012 with no changes except an annual update to the license header copyright year.  So you run `license:format` to update the range from 2012-2021 to 2012-2022.   You commit.  It complains about the header, because 10 commits ago for this file is the 2013 update, so you run format and it updates to 2013-2022.  You commit. It complains about the header, because 10 commits ago for this file is the 2014 update, so you run format and it updates to 2014-2022.  You commit... and so on 10 times until you're at "Copyright 2022".  For a file written in 2012 with no changes except the copyright header.

I'm happy to document this but I don't know where to put it.
 - does it belong in the main plugin as a parameter? It's only relevant when the git dependency is included.
 - it should probably be mentioned where the copyright year substitutions are documented but I don't see where that is.

dbwiddis:
macOS (native):
with default 10: 21.5 seconds.
with 100000: 26.5 seconds.

dbwiddis:
Linux (native M1 chip):
with default 10: 10.2 seconds
with 100000: 13.1 seconds

dbwiddis:
Solaris SPARC (shared compile farm machine):
with default 10: 1m49s
with 100000: 2m14s

dbwiddis:
AIX POWER8 (shared compile farm machine):
with default 10: 1m06s
with 100000: 1m09s

dbwiddis:
> I'm happy to document this but I don't know where to put it.

I found the docs in the README for the git dependency and updated it.

mathieucarbou:
@dbwiddis : thanks a lot for this analysis this is very helpful! 

If the property name makes sense (`license.git.copyrightLastYearMaxCommitsLookup`) we could stick with it, even if it is long. What I don't like in this name is the part `copyrightLastYear`... This parameter seems to be much more a global git parameter for maximum history lookup. So if you want to rename it to something shorter, but still keep an alias for the deprecated property so that it is backward compatible, I fully agree. ;-)

dbwiddis:
> If the property name makes sense (`license.git.copyrightLastYearMaxCommitsLookup`) we could stick with it, even if it is long. What I don't like in this name is the part `copyrightLastYear`... This parameter seems to be much more a global git parameter for maximum history lookup. So if you want to rename it to something shorter, but still keep an alias for the deprecated property so that it is backward compatible, I fully agree. ;-)

Well the property is only useful for the "last year" lookup but impacts all history.  I changed the name but kept the old value internally for compatibility. Should I add a log warning when the previous name is used?

mathieucarbou (approved):

mathieucarbou:
> Should I add a log warning when the previous name is used?

No, don't bother. Goal here would be to "hide" the old prop completely and only document the new one, but still keeping backward compat'.

Thanks a lot for the PR! Tremendous job and cleanup work! 👍

PR #322, opened 2022-03-25
[java] Wrap if statements in brackets

hazendaz:


mathieucarbou (approved):
Thanks! Completely agree 👍 

I didn't put the checkstyle and findbugs plugin in this project, but ideally we should.

Now that you did all this cleanup, it should be possible without a lot of impact.

PR #323, opened 2022-03-26
Skip existing copyright lines

### LINKED ISSUE #309: Keep existing copyright lines, use template for default
Software copyright statements are not always as consistent across all files as this plugin assumes, and there is no facility provided to permit different years or different authors other than the few provided substitutions (inception, file creation, latest change).

In order to use this plugin, maintainers are forced to adopt some text that applies to everyone, e.g., "Copyright 2020 the original author or authors" or in one project I maintain, a link to the Github "contributors" page.  But in another project I maintain, the convention is for the original author of a file to provide the copyright, e.g., one file might have
> Copyright 2012 Author Name

And another file might have
> Copyright 2015 Different Author

And perhaps on a third file
> Copyright 2012, 2016 Author Name and Different Author

(All with the boilerplate license text surrounding the copyright statement.)

A project which chooses to use the above format/policy has no way to effectively use this plugin.  However, there is no need to be so strict.  [License header templates](https://github.com/mathieucarbou/license-maven-plugin/tree/master/license-maven-plugin/src/main/resources/com/mycila/maven/plugin/license/templates) all use the general format:
* Stuff
* Copyright © year(s) owner(s)
* Other stuff

Currently the plugin has special treatment for the first and last lines, and there's an open PR (#243) to handle middle lines that could, in theory and with a lot of effort, be configured to handle this use case.  However, given the commonality across all licence headers (the copyright line is always on its own line) it should be easy to identify just that line using a well known regex (the word Copyright in between carriage returns).  The document and the header template could both be tokenized to compare the 3 "stuff" categories shown above, with the following caveat:  When no header exists at all, the template would fill in the copyright line; when a header exists containing the word Copyright in between CRs, it would be left completely alone.

I'm willing to submit a PR for this if there is interest. 
1. One solution would be similar to my draft PR #286, except instead of just skipping the first year, it would skip the entire line.
  * If no copyright line exists, would specify what to put in, e.g., current year, file creation year/author, etc.
  * If a copyright line exists, would totally ignore it
2. Another solution would be to add new substitutions (or a prefix before a substitution) for both year(s) and author(s) in which:
  * If no copyright line exists, would specify what to put in, e.g., current year, file creation year/author, etc.
  * If a copyright line exists, would append year(s) and name(s) if needed to existing

Let me know which you'd prefer, if there is interest.

dbwiddis:
Fixes #309 

Draft for review; still need to write tests

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
@@ -366,6 +366,12 @@ public abstract class AbstractLicenseMojo extends AbstractMojo {
   public boolean skipExistingHeaders = false;
 
   /**
+   * Skip the formatting of files in which the header matches other than text on the copyright line.
+   */
+  @Parameter(property = "license.skipExistingCopyrightLine", defaultValue = "false")
+  public boolean skipExistingCopyrightLine = false;
+
+  /**
    * When enforcing licenses on dependencies, exclude all but these scopes.
    */
   @Parameter(property = "license.dependencies.scope", required = true, defaultValue = "runtime")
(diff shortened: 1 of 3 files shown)


hazendaz:
keep open.

PR #326, opened 2022-03-27
Restore test runs, add jacoco, and fix resulting broken tests

hazendaz:
This resolves shortcoming in junit 5 migration.  I had initially wrote that code months ago and coming back to it only ran on command line.  Outside of missing import changes in a few spots at that time, it appeared my work was done.  However, the jump to junit 5 resulted in command line tests no longer running which then highlighted this was not complete on a small number of tests.  To ensure this type of thing never happens again, I've added jacoco to the mix.  Each pom has jacoco coverage number set.  The top one which is just pom is set to 100% for clarity reasons.  The other three are set to their appropriate current code coverage (git module 81%, svn module has no tests, and main plugin at 77%).  The svn module didn't actually need the setting as jacoco is skipped when no tests but presume we want to provision tests there at some point.  We may want to go further here and add coveralls to show the coverage overall for the repo itself.

note: jacoco reports will get added, these are not shown from main site but that site needs some retooling to fluido and few other things so it looks modern rather than legacy apache LAF. [update: noticed this behavior locally to not show jacoco from the site directly was due to attach of descriptor only rather than full site run that is going to happen after merge with github action.  I have updated to fluido locally though and it looks improved and also noting that main site on readme points to the old gh pages unless there is some other magic].

### DIFF
--- a/license-maven-plugin-git/pom.xml
+++ b/license-maven-plugin-git/pom.xml
@@ -39,6 +39,10 @@
     </site>
   </distributionManagement>
 
+  <properties>
+    <jacoco.minimum.coverage>0.81</jacoco.minimum.coverage>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>com.mycila</groupId>
--- a/license-maven-plugin-git/pom.xml
+++ b/license-maven-plugin-git/pom.xml
@@ -40,7 +40,7 @@
   </distributionManagement>
 
   <properties>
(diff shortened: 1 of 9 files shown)


mathieucarbou:
thanks! let's merge. there is an automated github action that generates the snapshot site at each commit: https://mycila.carbou.me/license-maven-plugin/reports/4.2-SNAPSHOT/

mathieucarbou (approved):

PR #327, opened 2022-03-27
Add feature to ignore git commits by id

philippn:
Hey @mathieucarbou ,

here is the PR, thanks for your consideration.

Kind regards, 
Philipp

### DIFF
--- a/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
+++ b/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
@@ -21,16 +21,21 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Arrays;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
+
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.Status;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.diff.DiffConfig;
 import org.eclipse.jgit.lib.Constants;
+import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.lib.PersonIdent;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.FollowFilter;
@@ -57,6 +62,7 @@ public class GitLookup implements Closeable {
   private static final String COPYRIGHT_LAST_YEAR_MAX_COMMITS_LOOKUP_KEY = "license.git.copyrightLastYearMaxCommitsLookup";
   public static final String COPYRIGHT_LAST_YEAR_SOURCE_KEY = "license.git.copyrightLastYearSource";
   public static final String COPYRIGHT_LAST_YEAR_TIME_ZONE_KEY = "license.git.copyrightLastYearTimeZone";
+  public static final String COMMITS_TO_IGNORE_KEY = "license.git.commitsToIgnore";
 
   public enum DateSource {
     AUTHOR, COMMITER
@@ -68,6 +74,7 @@ public class GitLookup implements Closeable {
   private final Repository repository;
   private final TimeZone timeZone;
   private final boolean shallow;
+  private final Set<ObjectId> commitsToIgnore;
 
   /**
    * Lazily initializes #gitLookup assuming that all subsequent calls to this method will be related
@@ -81,7 +88,7 @@ public class GitLookup implements Closeable {
     GitLookup.DateSource dateSource = GitLookup.DateSource.valueOf(
         dateSourceString.toUpperCase(Locale.US));
     String checkCommitsCountString = props.get(MAX_COMMITS_LOOKUP_KEY);
-    // Backwads compatibility
+    // Backwards compatibility
     if (checkCommitsCountString == null) {
       checkCommitsCountString = props.get(COPYRIGHT_LAST_YEAR_MAX_COMMITS_LOOKUP_KEY);
     }
@@ -89,6 +96,15 @@ public class GitLookup implements Closeable {
     if (checkCommitsCountString != null) {
       checkCommitsCountString = checkCommitsCountString.trim();
       checkCommitsCount = Integer.parseInt(checkCommitsCountString);
+    }
+    String commitsToIgnoreString = props.get(COMMITS_TO_IGNORE_KEY);
+    Set<ObjectId> commitsToIgnore = Collections.emptySet();
+    if (commitsToIgnoreString != null) {
+      commitsToIgnoreString = commitsToIgnoreString.trim();
+      commitsToIgnore = Arrays.stream(commitsToIgnoreString.split(","))
+              .map(String::trim)
+              .map(ObjectId::fromString)
+              .collect(Collectors.toSet());
     }
     final TimeZone timeZone;
     String tzString = props.get(COPYRIGHT_LAST_YEAR_TIME_ZONE_KEY);
@@ -108,7 +124,7 @@ public class GitLookup implements Closeable {
         throw new IllegalStateException(
             "Unexpected " + GitLookup.DateSource.class.getName() + " " + dateSource);
     }
-    return new GitLookup(file, dateSource, timeZone, checkCommitsCount);
+    return new GitLookup(file, dateSource, timeZone, checkCommitsCount, commitsToIgnore);
   }
 
   /**
@@ -122,10 +138,11 @@ public class GitLookup implements Closeable {
    * @param dateSource        where to read the commit dates from - committer date or author date
    * @param timeZone          the time zone if {@code dateSource} is {@link DateSource#COMMITER};
    *                          otherwise must be {@code null}.
-   * @param checkCommitsCount
+   * @param checkCommitsCount the number of historical commits, per file, to check
+   * @param commitsToIgnore   the commits to ignore while inspecting the history for {@code anyFile}
    * @throws IOException
    */
-  private GitLookup(File anyFile, DateSource dateSource, TimeZone timeZone, int checkCommitsCount) {
+  private GitLookup(File anyFile, DateSource dateSource, TimeZone timeZone, int checkCommitsCount, Set<ObjectId> commitsToIgnore) {
     try {
       this.repository = new FileRepositoryBuilder().findGitDir(anyFile).build();
       /* A workaround for  https://bugs.eclipse.org/bugs/show_bug.cgi?id=457961 */
@@ -155,6 +172,7 @@ public class GitLookup implements Closeable {
               "Unexpected " + DateSource.class.getName() + " " + dateSource);
       }
       this.checkCommitsCount = checkCommitsCount;
+      this.commitsToIgnore = commitsToIgnore;
     } catch (IOException e) {
       throw new UncheckedIOException(e);
     }
@@ -165,7 +183,7 @@ public class GitLookup implements Closeable {
    * year is taken either from the committer date or from the author identity depending on how {@link #dateSource} was
    * initialized.
    * <p>
-   * See also the note on time zones in {@link #GitLookup(File, DateSource, TimeZone, int)}.
+   * See also the note on time zones in {@link #GitLookup(File, DateSource, TimeZone, int, Set)}.
    *
    * @param file for which the year should be retrieved
    * @return year of last modification of the file
@@ -182,6 +200,9 @@ public class GitLookup implements Closeable {
     int commitYear = 0;
     RevWalk walk = getGitRevWalk(repoRelativePath, false);
     for (RevCommit commit : walk) {
+      if (commitsToIgnore.contains(commit.getId())) {
+        continue;
+      }
       int y = getYearFromCommit(commit);
       if (y > commitYear) {
         commitYear = y;
--- a/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
+++ b/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
@@ -15,21 +15,23 @@
  */
 package com.mycila.maven.plugin.license.git;
 
+import static java.util.Objects.requireNonNull;
+
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Arrays;
 import java.util.Calendar;
-import java.util.Collections;
 import java.util.Date;
 import java.util.Iterator;
-import java.util.Locale;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.stream.Collectors;
-
+import java.util.stream.Stream;
 import org.eclipse.jgit.api.Git;
 import org.eclipse.jgit.api.Status;
 import org.eclipse.jgit.api.errors.GitAPIException;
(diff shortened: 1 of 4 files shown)


philippn:
I also rebased it onto master just now

mathieucarbou:
@philippn : thanks a lot! Could you please also update the `README.md` in the git module to add a few lines to document this new feature ? There is a usage example that can be updated at least, plus eventually adding a comment line to explain the setting or a line below or above.

@dbwiddis @hazendaz : PR looks good to me as-is but please add your review if you can since you contribute a lot you might have your word on it too ;-)

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
```diff
@@ -34,6 +33,7 @@
   private static final String COPYRIGHT_LAST_YEAR_MAX_COMMITS_LOOKUP_KEY = "license.git.copyrightLastYearMaxCommitsLookup";
   public static final String COPYRIGHT_LAST_YEAR_SOURCE_KEY = "license.git.copyrightLastYearSource";
   public static final String COPYRIGHT_LAST_YEAR_TIME_ZONE_KEY = "license.git.copyrightLastYearTimeZone";
+  public static final String COMMITS_TO_IGNORE = "license.git.commitsToIgnore";
```
dbwiddis, line 36: This name should include `_KEY` in the name to distinguish it from the value.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -58,6 +54,7 @@
   private final Repository repository;
   private final TimeZone timeZone;
   private final boolean shallow;
+  private final Set<String> commitsToIgnore;
```
dbwiddis, line 57: Why store as a `String` rather than as a `RevCommit`?  You can use `fromString()` to convert.  This buys you:
 - Case insensitivity on the hex strings since you're converting to numeric values
 - Type safety/self-documenting code
 - Automatic format checking (e.g., invalid commit id strings will throw IllegalArgumentException and let the user know which commit ID is problematic). Otherwise you'd fail silently.
 - Less space. The strings take 80 bytes, the `RevCommit` stores in a 20-byte array.
 - There's already an `equals()` based on the numeric value so you can compare the `RevCommit` directly (`commitsToIgnore.contains(commit)`) as you iterate rather than filling up the heap using `.name()` to create a temporary `String` for every commit, for every file, while iterating.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -69,9 +66,10 @@
    * @param dateSource        where to read the commit dates from - committer date or author date
    * @param timeZone          the time zone if {@code dateSource} is {@link DateSource#COMMITER}; otherwise must be {@code null}.
    * @param checkCommitsCount
+   * @param commitsToIgnore
```
dbwiddis, line 69: Probably should explain what is expected in this argument.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
```diff
@@ -17,9 +17,8 @@
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Locale;
-import java.util.Properties;
-import java.util.TimeZone;
+import java.util.*;
```
dbwiddis, line 20: See above comment on wildcards.

license-maven-plugin-git/src/test/java/com/mycila/maven/plugin/license/git/GitLookupTest.java
```diff
@@ -159,7 +161,7 @@ public void reuseProvider() throws GitAPIException, IOException {
   @Test
   public void timezone() throws GitAPIException, IOException {
     try {
-      new GitLookup(gitRepoRoot.toFile(), DateSource.AUTHOR, TimeZone.getTimeZone("GMT"), 10);
+      new GitLookup(gitRepoRoot.toFile(), DateSource.AUTHOR, TimeZone.getTimeZone("GMT"), 10, emptySet());
```
dbwiddis, line 164: Unnecessary (as are the other changes in this file) if you use an overloaded constructor.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -95,14 +93,15 @@ public GitLookup(File anyFile, DateSource dateSource, TimeZone timeZone, int che
...
+    this.commitsToIgnore = commitsToIgnore;
   }
 
   /**
    * Returns the year of the last change of the given {@code file} based on the history of the present git branch. The
    * year is taken either from the committer date or from the author identity depending on how {@link #dateSource} was
    * initialized.
    * <p>
-   * See also the note on time zones in {@link #GitLookup(File, DateSource, TimeZone, int)}.
+   * See also the note on time zones in {@link #GitLookup(File, DateSource, TimeZone, int, Set)}.
```
dbwiddis, line 104: This doesn't need to change if you use an overload. Also, interestingly, there is no note on time zones. 😁

dbwiddis (changes requested):
Comments inline. 
 - [x] Also you should add a test to `GitLookupTest` to verify that it works.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -34,11 +34,7 @@
 
 import java.io.File;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Calendar;
-import java.util.Date;
-import java.util.Iterator;
-import java.util.TimeZone;
+import java.util.*;
```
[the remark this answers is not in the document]
mathieucarbou: @dbwiddis : I think the priority could be to add the checkstyle plugin to the project in order to help contributors :-)
Some IDE are configured by default to use wildcards when optimizing imports.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
```diff
@@ -57,13 +57,20 @@ GitLookup getGitLookup(AbstractLicenseMojo mojo, File file, Properties props) th
...
           int checkCommitsCount = Integer.MAX_VALUE;
           if (checkCommitsCountString != null) {
             checkCommitsCountString = checkCommitsCountString.trim();
             checkCommitsCount = Integer.parseInt(checkCommitsCountString);
           }
+          String commitsToIgnoreString = props.getProperty(COMMITS_TO_IGNORE);
+          Set<String> commitsToIgnore = Collections.emptySet();
+          if (commitsToIgnoreString != null) {
+            commitsToIgnoreString = commitsToIgnoreString.trim();
+            commitsToIgnore = Arrays.stream(commitsToIgnoreString.split(","))
```
dbwiddis, line 71: One other suggestion.  Splitting on a comma forces users into a long one-line string here.  You could split on a non-hex digit, e.g., `[^0-9a-fA-F]+` (and also filter out empty strings after the split) which would allow flexibility such as listing commits on newlines like:
```xml
<license.git.commitsToIgnore>
  6eb93dcfc014bfd3980c1f1c43f10fd59422107e,
  ad1e59a6afc5a1a7edbdc8bcf949bb2f8682f074,
  70403fdbc424e5d2045389e685f9bb9bd696d5cd
</license.git.commitsToIgnore>
```

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -69,9 +66,10 @@
    * @param dateSource        where to read the commit dates from - committer date or author date
    * @param timeZone          the time zone if {@code dateSource} is {@link DateSource#COMMITER}; otherwise must be {@code null}.
    * @param checkCommitsCount
+   * @param commitsToIgnore
    * @throws IOException
    */
-  public GitLookup(File anyFile, DateSource dateSource, TimeZone timeZone, int checkCommitsCount) throws IOException {
```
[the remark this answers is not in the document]
dbwiddis: Looking again, this is fine as-is.  Other than in tests, `GitLookup` is set up to be used by the `GitPropertiesProvider` with lazy initialization of a static variable, so there really should be only one entry point.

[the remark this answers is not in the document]
mathieucarbou: Let's @philippn decide because both ways are debatable and ok IMO.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
dbwiddis: By the way, the reason I'm making this suggestion is that I really think this is a cool feature and I'm eager to use it, and somewhat biased by how I want to use it. ;)

philippn:
I'm a bit overwhelmed by all this feedback, please bear with me :-)

So in the order of the points being mentioned:

- I added a paragraph to the readme.md
- I added myself to the contributors, thanks for pointing that out
- I added two new tests to `GitLookupTest`; no here is actually where things start to get interesting. On my machine (Windows, IntelliJ) the tests weren't picked up, neither by IntelliJ itself nor by Maven. Dunno whats going on here but changing line 60 to `gitRepoRoot = unzipDestination.resolve("git-test-repo");` solved the issue for me, at least it it then running in IntelliJ.

Regarding the concrete things pointed out by @dbwiddis I will comment on each individually.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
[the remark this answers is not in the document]
philippn: Roger, I changed that back

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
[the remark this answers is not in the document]
philippn: Since it is really only used at one place, lets leave it that way.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
philippn: Roger, thanks for pointing that out

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
philippn: Nice catch, I have changed it accordingly.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
philippn: Roger

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
philippn: Roger, I changed that back

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
philippn: I looked into this, I'm a bit vary to use lots of regex magic here as it is not easily testable. It would probably make sense to create a test for `GitPropertiesProvider`, but that would require factoring out the instantiation of the `GitLookup` into some kind of factory.

So for the time being, I added a `trim` operation to the streaming expression, that should basically result in the same behaviour. So technically, I believe your example above should work :-)

dbwiddis: > I looked into this, I'm a bit vary to use lots of regex magic here as it is not easily testable. 

Fair enough. I hit regex101 to play around with my proposal and it took a few iterations to confirm what I expected; and the added complexity of producing empty strings that you then have to filter was undesirable. I like your `trim()` solution and it (should) work with my use case.

> It would probably make sense to create a test for `GitPropertiesProvider`, but that would require factoring out the instantiation of the `GitLookup` into some kind of factory.

Yeah. I'm not fond of the existing lazy initialization but haven't had time to think of a better way to do it.

dbwiddis (approved):

dbwiddis:
> I'm a bit overwhelmed by all this feedback, please bear with me :-)

Speaking for project maintainers in general, we are thrilled with contributions from the community and we will happily bear with you as needed.  Thanks for contributing and being so gracious with the feedback.

This LGTM at this point.   Not sure what's going on with the tests, but if @mathieucarbou wants to merge this push out an rc3 that I can fiddle with real-life testing on my own project I'm happy to do that!

dbwiddis:
@hazendaz JDK 17 / 18 tests are failing with "Not executing Javadoc as the project is not a Java classpath-capable package" ... hope that's something you're already working on. Do we need to rebase this PR?

hazendaz:
> are failing with "Not executing Javadoc as the project is not a Java classpath-capable package" ... hope that's something you're already working on. Do we need to rebase this PR?

Had not seen that yet but will take a look at that one.  Thanks.

hazendaz:
@dbwiddis javadoc may be unrelated issue.  The issue that jump sout at failure is test coverage drops down by 1% on 17/18.  Not sure exactly why but I had the same issue when setting jacoco up there.  Typically I go with more strict enforcement of coverage but in this case wanted to go with the closest to IDE coverage stats in Eclipse.  So to fix that @philippn, its easy go to the 2 submodule poms (license-maven-plugin and license-maven-plugin-git.  Just drop the jacoco.minimum coverage number by 1% in both.  That should get it to green here.

mathieucarbou (dismissed):
Please wait for #339 to be merged first and rebase onto after.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitPropertiesProvider.java
mathieucarbou: > > It would probably make sense to create a test for `GitPropertiesProvider`, but that would require factoring out the instantiation of the `GitLookup` into some kind of factory.
> 
> Yeah. I'm not fond of the existing lazy initialization but haven't had time to think of a better way to do it.

So that is fixed in #339

mathieucarbou:
> Please wait for #339 to be merged first and rebase onto after.

#339 is merged!

dbwiddis (approved):

mathieucarbou:
@dbwiddis : your merge commit looks weird ;-)
The correct way for this PR is to rebase the original commit on top of master, solve conflicts, and merge it with a merge commit.
Don't know if @philippn has the time for that, or otherwise if not we rebase and force push his commit if his branch his opened, or create a new PR.

mathieucarbou (changes requested):
PR has to be rebased on top of master and conflicts solved as part of the rebase.

dbwiddis:
> @dbwiddis : your merge commit looks weird ;-) 

Github commandline ain't all it's cracked up to be.

Unfortunately the whole #339 refactor made resolving conflicts nontrivial.  Will try again.

mathieucarbou:
> Don't ask why there are two merge commits. But I've fixed conflicts 4 times now and I think it's correct now. 😁

I'll check. Not normal... gît fetch upstream && gît rebase -i upstream/master then fix conflits during rebase, there should be one commit and no merge commit

dbwiddis:
Yeah, I know how it's *supposed* to work.  Between the GitHub CLI and my IDE things didn't work out well the first time.

I successfully merged the second go around but then ended up unable to push back to this branch, so I cherry-picked the merge commit onto this branch and it seemed to work, with the exception of showing two commits.

mathieucarbou:
> 

That is normal that you were not able to push back because rebasing changes the history and commit ID to place it on top of master. You have to force the puah, which will rewrite the branch history for this commit and GitHub will show your icon next to @philippn 's icon

dbwiddis:
Well, in any case, the two commits are identical and the result of all this is correctly working code that can be merged.   If someone wants to put in the work to back those commits out and start afresh, feel free.  Having done the same changes 4 times, I'm done. :grin:

dbwiddis:
Well, that worked just fine at home on my mac.  😏

mathieucarbou:
> Well, that worked just fine at home on my mac. smirk

Still not rebased on master if we look at the graph: https://github.com/mathieucarbou/license-maven-plugin/network

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -90,6 +97,15 @@ public static GitLookup create(File file, Map<String, String> props) {
...
       checkCommitsCount = Integer.parseInt(checkCommitsCountString);
     }
+    String commitsToIgnoreString = props.get(COMMITS_TO_IGNORE_KEY);
+    Set<ObjectId> commitsToIgnore = Collections.emptySet();
+    if (commitsToIgnoreString != null) {
+      commitsToIgnoreString = commitsToIgnoreString.trim();
+      commitsToIgnore = Arrays.stream(commitsToIgnoreString.split(","))
+              .map(String::trim)
+              .map(ObjectId::fromString)
+              .collect(Collectors.toSet());
```
mathieucarbou, line 107: there is a bug here: we need to filter out empty strings after the split

mathieucarbou (changes requested):

mathieucarbou (approved):

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
mathieucarbou: @dbwiddis @philippn : FYI added a second commit to rework that part

PR #331, opened 2022-03-27
[actions] Go to release jdk 18 and add 19-ea jdk

hazendaz:


hazendaz:
closing for now, will re-open with just jdk 18 update.  jdk 19 while fine is not yet supported by groovy so we need to wait a bit for that.

PR #332, opened 2022-03-27
[pom] Add missing maven plugin annotations to clear warnings

hazendaz:


hazendaz:
warning in this case is that maven complains it doesn't understand the enums used as missing the class.  Adding this artifact resolves the warning.

dbwiddis (approved):

hazendaz:
note on provided here:  maven as a whole is going towards provided on these so they don't pull many maven copies per build.  Realistically, getting compliant here at least for this build doesn't really overall help at least currently.  Too many plugins still bring in whatever maven they are compiled against so lots of 2 and 3 versions show up.  This does however fix the flip end where usage of this plugin will not pull these to the .m2 cache.

PR #333, opened 2022-03-27
[pom] Override poms from parent to latest versions

hazendaz:
realized tests were failing - override all the out of date plugins to ensure we are running as new as possible.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,36 @@
       <plugins>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


dbwiddis:
Is dependabot turned on here? If not, it should be...

hazendaz:
@dbwiddis It is.  These are in the parent pom this project uses, not sure where that code is exactly.  Its out of date.  Proper way would be to fix there but there are a number of overrides already so just did them all.

dbwiddis (approved):

PR #335, opened 2022-03-28
Fix unclosed reader (reported by SonarLint)

philippn:
Hey @mathieucarbou 

while looking at the source code for the other PR, I stumbled about an issue reported by SonarLint, please find attached PR.

Thanks,
Philipp

### DIFF
--- a/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
+++ b/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
@@ -20,6 +20,7 @@ import org.eclipse.jgit.api.Status;
 import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.diff.DiffConfig;
 import org.eclipse.jgit.lib.Constants;
+import org.eclipse.jgit.lib.ObjectReader;
 import org.eclipse.jgit.lib.PersonIdent;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.revwalk.FollowFilter;
@@ -76,8 +77,9 @@ public class GitLookup {
     this.repository = new FileRepositoryBuilder().findGitDir(anyFile).build();
     /* A workaround for  https://bugs.eclipse.org/bugs/show_bug.cgi?id=457961 */
     // Also contains contents of .git/shallow and can detect shallow repo
-    this.shallow = !this.repository.getObjectDatabase().newReader().getShallowCommits().isEmpty();
-
+    try (ObjectReader objectReader = this.repository.getObjectDatabase().newReader()) {
+      this.shallow = !objectReader.getShallowCommits().isEmpty();
+    }
     this.pathResolver = new GitPathResolver(repository.getWorkTree().getAbsolutePath());
     this.dateSource = dateSource;
     switch (dateSource) {


dbwiddis:
(Quick reply) I'm pretty sure I hit the same sonarlint issue and dug into this and it's not a problem.  Will re-trace my steps and produce a longer reply later.

hazendaz:
@dbwiddis This is valid.  This and another you flagged in a prior commit as to be ignored.  Jgit specifically states that its not closed when opened in their javadoc and that user must close it.  If try with resources was used but not needed as not closable, the code would actually error out.

hazendaz:
of course I need to find that now as not seeing the javadoc that stated that at the moment but looking.  anyways, I have same change on this one locally as well as one more in same class.  Additionally the jgit code opens tries around the same code internally.

hazendaz:
ok found the other one, around line 195 on the new Git, the javadoc is rather explicit there.  

```
The caller is responsible for closing the repository;
```
While it doesn't state that on this one, the internals that call the same way do in fact uses try with resources in jgit so its needed.  Further looking at the one that additionally needs it, I'm pretty sure you added the '@SuppressWarnings("resource")'.  I meant to look into why as it is very firm on the need there.  Sonarlint in both cases will flag both - well if the second one has the ignore removed.

dbwiddis:
I know what you're saying and read the same docs and was this close --->| |<--- to a similar report.  I'm still pretty sure I dug enough to find out that it wasn't. I might be misremembering, or have made a wrong assumption somewhere, of course.  Are we at the "wager a drink of choice" stage on this yet? :)

hazendaz:
@dbwiddis Maybe :)  I've seen sonarlint state try with resources needed in the past but class is not closable type.  These are so I suspect its right.  Given, at least on the one that jgit doesn't really say, has code in same class doing the exact same thing, going to trust they know what to do :)  I've actually tested this over on waffle project that has a good level of classes and at least from aspect of license plugin 4.1 to 4.2.rc2 and then the latest snapshot from my local, the speeds are improving give or take 30 seconds - not a good measure of test, just many maven clean install runs.  So it seems at least harmless and at very least another contributor coming to same conclusion.  I'm ok with the change, one less off my outstanding ;)

dbwiddis:
> So it seems at least harmless and at very least another contributor coming to same conclusion. 

I'll agree that it's harmless.  And as regards avoiding other people deep diving as much as I did (or me doing it again to justify why it's not necessary) I'm not overly opposed to this PR.

I'm just Type-A enough to, at some point in the next few hours, digging in to this again for fun.

hazendaz:
ok will let you mull over this and when you are happy you can merge it assuming you agree :)

dbwiddis:
So, retraced my steps and I'm not sure I ever found that it was *not* needed.  I think I got to the point to where I wasn't sure it wouldn't break anything.   Following the URL in the comment leads to this comment (from 2015), showing a symptom using this plugin: https://bugs.eclipse.org/bugs/show_bug.cgi?id=457961#c5 and the line (up to the `getShallowCommits()` portion) used as a workaround.

I think I was of the opinion that whatever it was doing needed to stay initialized to do its magic.

I'm still firmly on the fence on this; unless I know why it's needed as a workaround, I don't know if "closing" it will undo the workaround.

mathieucarbou:
I think this workaround has been introduced by @ppalaga and is still needed until fixed in JGit. 

The line `repo.getObjectDatabase().newReader().getShallowCommits();` seems to be required as per the comment says to 
`// Force initialization of shallow commits to avoid later concurrent modification issue`.

The implementation of `getShallowCommits()` reads and caches. So it has to be called before doing any RevWalk.

`.newReader().getShallowCommits()` calls `FileObjectDatabase#getShallowCommits`, which is the object referenced by the repository itself.

What's important is to close the repository, which is the object holding the DB and its cache. Which isn't done atm...

mathieucarbou:
will move to draft atm

mathieucarbou:
Closing this: initialization and close of providers are now fixed as part of #339

PR #337, opened 2022-03-28
[pom] Add checkstyle with last jdk < 11 usage at 9.3 and basic setup

hazendaz:
The checkstyle file is identical to google's from the checkstyle jar with this project license header added as at least at the moment was required but probably should be ignored.

Reason for using the file directly is that there is some expectation of changes needed.  I forsee at least 100 characters per line being set to 120 for more reasonable usage.  There could be others so starting there helps.  Do note that means that with any update on checkstyle, given they do not use semantic versioning there is little backwards compatibility even at minor revisions.  Therefore, its entirely possible it fails fast when updated as it does with 10.0 already.  Thus for any upgrades, its important to extract the content and overlay it reapplying any custom changes that occur later.  Typically that is made easier by at least wrapping begin/end sections around what we might envision as a google rules + our own rules such as line length.  To that end, maybe our license additionall here is not really a concern either.  We retain the original so its compliant anyways and us marking it indicates we might have some custom usage.

I did stick this under the .mvn folder given its build related and more specifically attached to the maven plugin for checkstyle.  That may or may not be a concern if we use other tools later like 'rewrite' that can auto update some static code for us including what we require for checkstyle.

Mainly added @dbwiddis to review given original request for feature.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -308,6 +308,34 @@ limitations under the License.
         </configuration>
       </plugin>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>3.1.2</version>
+        <configuration>
+            <configLocation>.mvn/checkstyle.xml</configLocation>
+            <encoding>UTF-8</encoding>
+            <consoleOutput>true</consoleOutput>
+            <failsOnError>true</failsOnError>
+            <includeTestSourceDirectory>true</includeTestSourceDirectory>
+        </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>com.puppycrawl.tools</groupId>
+            <artifactId>checkstyle</artifactId>
+            <version>9.3</version>
+          </dependency>
+        </dependencies>
+        <executions>
+          <execution>
+            <id>validate</id>
+            <phase>validate</phase>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
         <groupId>org.jacoco</groupId>
         <artifactId>jacoco-maven-plugin</artifactId>
         <version>0.8.7</version>
--- a/.mvn/checkstyle.xml
+++ b/.mvn/checkstyle.xml
@@ -1,0 +1,381 @@
+<?xml version="1.0"?>
+<!--
+
+    Copyright (C) 2008-2022 Mycila (mathieu.carbou@gmail.com)
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+            http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


hazendaz:
looking at failure now.

hazendaz:
failure issue resolved.  at the moment, the copyright year is still 2021 on this project.  I had this at 2022 on top of fixing that so I had to adjust back to 2021.  It was not clear on the error which is probably another thing we need to look more into and may have already been noted.

dbwiddis (approved):

mathieucarbou:
I think we have a pretty huge bug...

I have changed `<year>${project.inceptionYear}-2021</year>` to `<year>${project.inceptionYear}-2022</year>` and `/mvnw license:check` still works and `./mvnw license:format` does not reformat anything.

pom.xml
```diff
@@ -307,6 +307,34 @@ limitations under the License.
...
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>3.1.2</version>
+        <configuration>
+            <configLocation>.mvn/checkstyle.xml</configLocation>
+            <encoding>UTF-8</encoding>
+            <consoleOutput>true</consoleOutput>
+            <failsOnError>true</failsOnError>
+            <linkXRef>false</linkXRef>
```
dbwiddis, line 319: Why do we not want the linkXRef?

pom.xml
```diff
@@ -307,6 +307,34 @@ limitations under the License.
...
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>3.1.2</version>
+        <configuration>
+            <configLocation>.mvn/checkstyle.xml</configLocation>
+            <encoding>UTF-8</encoding>
+            <consoleOutput>true</consoleOutput>
+            <failsOnError>true</failsOnError>
```
dbwiddis, line 318: Consider adding `<includeTestSourceDirectory>true</includeTestSourceDirectory>`

pom.xml
```diff
@@ -307,6 +307,34 @@ limitations under the License.
...
+            <linkXRef>false</linkXRef>
+        </configuration>
+        <dependencies>
+          <dependency>
+            <groupId>com.puppycrawl.tools</groupId>
+            <artifactId>checkstyle</artifactId>
+            <version>9.3</version>
```
dbwiddis, line 325: Latest is 10.1

dbwiddis (approved):
LGTM. Suggestions inline.

pom.xml
dbwiddis: Ah but I read the subject line of this PR and see this was intentional. 😁

mathieucarbou:
@dbwiddis @hazendaz : let's get this PR rebased and ready to be merged 👍

pom.xml
hazendaz: will look at this with rebase.  likely will remove it once I see what its purpose was.  thanks.

mathieucarbou:
@dbwiddis @hazendaz : crap! merged the wrong PR sorry!
I have reverted... It needs to be re-opened

PR #342, opened 2022-03-28
Also commit new generated files

mathieucarbou:


mathieucarbou:
merging directly to try teh change. The generated PR for the snapshot site should contain new files related to JaCoCo

PR #437, opened 2022-12-24
[leaks] Fix resource leaks

hazendaz:
jgit is clear in javadoc to do this as well as being caught.  One was being suppressed, now both are handled so no more warnings.


mathieucarbou (approved):
I bumped into this already and checked the JGit impl: there is no leak because of the repository impl that is used. Plus JGit is closed by PropertiesProvider#close().

Though I completely agree this is bad to not properly handle try-resources ;-)

PR #441, opened 2023-01-01
Update plugin version for license-maven-plugin-git README

tisonkun:
See #440.


hazendaz:
approving to merge, but note: never use readme's to indicate obsolete versions especially when it states the same there already.  2.x would not at all be supported, its 8 years old.  its important to look to either maven central or releases stated as released per the tags to know what is supported which is always pretty much the latest.  RC here normally would be bad but given its been RC for the longest time, its acceptable as its been used heavily at that revision otherwise.

hazendaz (approved):

PR #442, opened 2023-01-01
Accept skipLinePattern as property name

tisonkun:
This should be intuitive. I configure `skipLinePattern` and it doesn't work. Later I found it is `skipLine` which is weird.


hazendaz:
@mathieucarbou There is probably more to this as an issue than just this fix.  I'm not familiar with the usage here but skipLine and skipLinePattern seems all over the code.  Not entirely sure if this is all that is needed.

mathieucarbou:
I will have a look.

mathieucarbou (changes requested):
Even if I agree with the new proposed name, the PR is incomplete:

- other places in the code has to be changed
- the schema needs to be updated to reflect the new name and deprecated one still hasto be supported
- a rename of the different variables in the code needs to happen

tisonkun:
Thanks for your explanation! Then I'm going to close this PR as I _can_ accept the current solution and don't have too much time to do the whole refactoring.

PR #483, opened 2023-02-13
[java] Cleanup isAllowed method as it can be much more simple

hazendaz:


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/dependencies/LicensePolicy.java
```diff
@@ -51,16 +51,10 @@ public boolean getPredicate() {
...
      */
     public boolean isAllowed(final boolean matched) {
-      if (matched && allowed) {
-        return true;
-      } else if (!matched && !allowed) {
-        return true;
-      } else {
-        return false;
-      }
+      return (matched && allowed) || (!matched && !allowed);
```
mathieucarbou, line 57: `return !(matched ^ allowed)`

mathieucarbou (changes requested):

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/dependencies/LicensePolicy.java
hazendaz: think that does not apply given how the checks are but Eclipse is suggesting far, far easier -> 'returned matched == allowed'.

hazendaz: updating PR with suggestion from eclipse better look at.

mathieucarbou: lol indeed ;-) `false == false`

mathieucarbou (approved):

PR #488, opened 2023-02-14
[tests] Remove duplicate imports from earlier change

hazendaz:


hazendaz:
mistake caused during parameterized test introduction as all junit entries placed together but the other original set was missed.

PR #489, opened 2023-02-14
Add plexus utils as compile scope, use non deprecated failsafe config, remove optional true on plugin annotations

hazendaz:


hazendaz:
Right now you won't see why plexus utils is used.  But you will see it if dependency tree is upgraded as all newer maven components no longer leak items unnecessarily.  This is more important for maven 3.9 and 4.0.  With maven 3.9 out and us now building on it, its critical to get this right.  For the other items just minor cleanup.

On the dependency tree note, I'll raise a separate PR upgrading that as it will need extra set of eyes to make sure its working as expected.  Specifically need to have Michael Simons verify since he was able to resolve other issues around same area.  For now this is harmless here and will help limit what he has to look at.

mathieucarbou (approved):

PR #494, opened 2023-02-15
Raise dependency tree to 3.2.1 and remove internal touching of maven

hazendaz:
internals no longer there that way.  probably not a good idea to touch their internals.

### DIFF
--- a/license-maven-plugin-git/pom.xml
+++ b/license-maven-plugin-git/pom.xml
@@ -45,10 +45,9 @@
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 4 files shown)


hazendaz:
@michael-simons  Any chance you could take a look at this?  Think it affects item you previously were fixing.  Not sure why we were touching maven internals but its different now and this was easiest path to fix.  It may not be the right fix so open to any suggestions or to understand if doing this breaks anything for you.

mathieucarbou (approved):

PR #533, opened 2023-04-01
Add fix guidance to default `check` error message

### LINKED ISSUE #531: Add suggestion to default `errorMessage` on how to fix missing/wrong license headers
**Describe the feature**

Currently, I define a `check` execution in my projects to make sure each file has the proper license header and I let the CI fail if something is missing. This gives immediate feedback in PRs but the default error message doesn't give indications on how to fix the problem.

Without guidance, it's assumed that the PR author is or needs to get familiar with the `license-maven-plugin`, which is not always the case, especially for one-time contributors.

The project-specific contributing guidelines could contain some details about the license headers and how to fix the missing ones but I believe it would be more effective to have guidance directly in the error message. For this reason, I propose to define the default error message like the following:

```
Some files do not have the expected license header. Run license:format to update them.
```

I'm happy to raise a PR if the idea is accepted.

**Any alternative you've considered ?**

To give direct feedback to users, I currently define a custom error message with the content above:

```xml
<configuration>
  <errorMessage>Some files do not have the expected license header. Run license:format to update them.</errorMessage>
</configuration>
```

scordio:
Closes #531.

This changes the `check` mojo default error message from:

```
Some files do not have the expected license header
```

to:

```
Some files do not have the expected license header. Run license:format to update them.
```

Two questions:
- [ ] Should I manually update the error message in the files under `docs/reports/4.3-SNAPSHOT`?
- [ ] If I execute `mvn clean verify`, some non-Java files in scope for git are modified. Am I supposed to push these changes as well?

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/LicenseCheckMojo.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/LicenseCheckMojo.java
@@ -44,8 +44,8 @@ import java.util.stream.Collectors;
 @Mojo(name = "check", defaultPhase = LifecyclePhase.VERIFY, threadSafe = true)
 public final class LicenseCheckMojo extends AbstractLicenseMojo {
 
-  @Parameter(property = "license.errorMessage", defaultValue = "Some files do not have the expected license header")
-  public String errorMessage = "Some files do not have the expected license header";
+  @Parameter(property = "license.errorMessage", defaultValue = "Some files do not have the expected license header. Run license:format to update them.")
+  public String errorMessage = "Some files do not have the expected license header. Run license:format to update them.";
 
   public final Collection<File> missingHeaders = new ConcurrentLinkedQueue<>();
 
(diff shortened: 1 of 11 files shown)


hazendaz (approved):

hazendaz:
@scordio On your two questions.

On both, the docs are auto generated.  I am not 100% on the first one needing something manual but @mathieucarbou can confirm.  On the second issue, very aware of that issue myself, just discard when you see that for now.  We do need to make that better.  In the past I used to commit those separately then never push as you may have as well, then I typically go away and don't come back for weeks or months then get very confused why those exist.  Its not a good contributor experience let alone new contributor experience.  If you have time to look into how we might make that not occur, awesome, if not, think enough of an answer here to address.

On the change, wow we have that in a lot of places, Pressure sure we can use some sort of static to write once rather than many, but keeping scope of the change as-is.  Thank you for both inquiring first to not waste your time and second for taking the time to contribute.  It is very much appreciated and hope to see more contributions.  The approach here I very much like as I noted on your original issue question regarding this.  Its extremely helpful to tell a user how to address anything rather than just error.  Recently even maven owned plugins are going a long way to make it more user friendly in that way.  Clearly when someone failed to do what was needed such as run the plugin, they need a boost.  There are other ways to automate such a condition such as a git hook but clearly getting all to install is yet another issue.  So this along is worth its weight in gold IMO.

hazendaz (approved):

hazendaz:
One other note just as an FYI, we did just release 4.2.  So we may be a bit before 4.3.  That could change though depending how fast we can address some long standing issues we have with slow performance in general when using svn or git extensions to do timebased checks.  Its also feasible that maybe we run a little faster than before but either way, it will probably be a few months if I had to guess before we release next cut so we have at least a good number of changes.

PR #564, opened 2023-05-27
[m2e] Tell m2e what it should do with this plugin (ie ignore during m2e executions)

hazendaz:


hazendaz:
This is necessary integration to m2e so that it knows what to do with the plugin executions.  In this case, it will tell m2e (ie Eclipse) to ignore the plugin.  We would not want it to generally do anything there as its so heavy hitting already and that would be smoothest approach.  If users want it to run in m2e though, they simply need to add the standard m2e integration in their pom and set to appropriate levels.  The main point here is so that m2e doesn't complain it cannot figure out the intent of this plugin.

mathieucarbou (approved):

PR #565, opened 2023-05-27
[GHA] Drop jdk 19, add 21-ea, and use 20 GA

hazendaz:


hazendaz:
Hang on this for the time being, need to look into integration tests to see what went wrong there for jdk 21 which isn't 100% clear here yet as not fully setup on invoker to tell us why.

hazendaz:
groovy setup was incorrect on invoker.  Now streaming errors to console if invoker fails and further overriding the correct groovy modules.  Invoker is still on groovy 3 and older.  We were potentially correctly overriding at one point but they had changed since to only use certain portions instead of all and so we were no longer overriding anything.  Now we are again.

mathieucarbou (approved):

PR #567, opened 2023-05-27
Add few quick tips to readme.md to avoid longer read through and add more exclusions

hazendaz:


hazendaz:
Exclusions added...

For certificate types, more recently added and most often manually setup on ignores now
- Add .cer - certificate
- Add .der - digital certificate

For some use cases, its possible projects may be importing ear and war file types too just like jars so this helps there or they have them checked in - very specific use case where no formal hosting exists.  This happens with many vendors supporting various industries where they don't have artifactory, sonatype, etc.
- Add .ear - ear archive
- Add .war - war archive

Newer bzip type
- Add .tar.bz3 - bzip 3

Many users continually ask to skip files and the list of files needed to skip is rather extensive at times with many cases no file extensions.  To easily resolve that, for 2 to 3 years, we have implemented the following which has greatly reduced need for downstream (non super pom) usage of license plugin to ignore wide area needed without any further configuration.
- Add new explicit location to ignore license all together. '**/unlicensed/**'


I did notice in reviewing the current exclusions we are still certainly missing quite a bit that should be excluded.  However, I don't have personal experience in those exclusions needs at the moment so I leaned only towards adding what I'm already overriding in my super pom at work where I'd like to cleanup and reduce more the configuration of the plugin to mostly defaults.

For the readme, this is for the lazy in all of us :)  Just want a quick where is this stuff hidden in the plugin.  I've been doing this in our super pom for years now and continually have to go there to find it.  For our users, they typically don't even look in the super pom and if pointed here at best they will find the defaults eventually but not the types as easily as looking directly at the code.  So this achieves quickly getting them there as they are more likely to go look at the github plugin.

mathieucarbou (approved):

PR #573, opened 2023-05-28
[pom] Add missing plugin annotatons to git/fs modules

hazendaz:
solves warnings in tests


hazendaz:
two modules were throwing warnings during build due to missing this in test scope.  Just left as provided as it is in the other modules to be similar setup which resolved the issues.

mathieucarbou (approved):

PR #575, opened 2023-05-28
[reporting] Do not run reports to docs in windows

hazendaz:
do this only on linux where its needed to occur.

Currently usage of this makes it hard for contributors as they are blasted with many changes on every run.  Intent here was to run this on GHA only thus fixing so that it no longer runs on windows.  Maybe there is a more friendly determination of GHA but it is set to use ubuntu there.


mathieucarbou (approved):

mathieucarbou:
Yeah I understand the right solution might lead to a bigger refactor and this workaround is completely acceptable ;-)

PR #579, opened 2023-05-29
[javadocs] Fix javadocs clearing all warnings

hazendaz:
Resolves all javadoc warnings to console that are a result of failure to comply with requested setup of same.

This does not resolve one left over item where the link is being redirected to jdk 20 but makes the output of the build finally in most reasonable state.

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/document/Document.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/document/Document.java
@@ -41,6 +41,15 @@ public final class Document {
   private HeaderParser parser;
 
 
+  /**
+   * Instantiates a new document.
+   *
+   * @param file the file
+   * @param headerDefinition the header definition
+   * @param encoding the encoding
+   * @param keywords the keywords
+   * @param documentPropertiesLoader the document properties loader
+   */
   public Document(File file, HeaderDefinition headerDefinition, String encoding, String[] keywords, DocumentPropertiesLoader documentPropertiesLoader) {
     this.keywords = keywords.clone();
     this.file = file;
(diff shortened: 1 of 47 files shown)


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/document/Document.java
```diff
@@ -49,26 +58,58 @@ public Document(File file, HeaderDefinition headerDefinition, String encoding, S
...
+   */
   public File getFile() {
     return file;
   }
 
+  /**
+   * Gets the file path.
+   *
+   * @return the file path
+   */
```
mathieucarbou, line 83: please remove all the javadoc not bringing any information.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/document/DocumentType.java
```diff
@@ -28,87 +28,243 @@
...
+  /** The ruby. */
   RUBY("rb", HeaderType.SCRIPT_STYLE),
+
+  /** The scala. */
   SCALA("scala", HeaderType.SLASHSTAR_STYLE),
+
+  /** The scaml. */
   SCAML("scaml", HeaderType.HAML_STYLE),
+
+  /** The scss. */
```
mathieucarbou, line 190: please don't add javadoc if you don't document.... This does not bring any information.

mathieucarbou (changes requested):

hazendaz:
converted to draft so these can be better written out to be useful.  the core reason was to remove the 100s of warnings at maven.  keeping in draft will avoid accidental merging until work can later be completed.

mathieucarbou:
> converted to draft so these can be better written out to be useful. the core reason was to remove the 100s of warnings at maven. keeping in draft will avoid accidental merging until work can later be completed.

I understand the reason, but that's not a good fix ;-) This is a plugin so I would suggest either we remove the javadoc from everywhere except the plugin metadata (and convert existing doc to internal doc with `//`), or we correctly add the javadoc. And if some warnings require to be ignored at the javadoc plugin itself, then let's ignore them, globally or per file.

hazendaz:
closing for now to prevent noise as I rebase branches locally on my end.  Will reopen once appropriate work done here.

PR #582, opened 2023-05-29
[actions] Add drop to jdk 8 and maven 3.2.5 to run integration tests after initial build completion

hazendaz:
- Build will run with respective Maven in wrapper and Java designated version first.
- Using the already built binary of license plugin, build will download lowest known to work maven version 3.2.5 using mvn, not the wrapper so maven version may vary here.  This is done this way because windows won't allow wrapper to download wrapper due to jar being open.
- Finally integration tests baked into build will run against java 8 and maven 3.2.5

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,10 +20,19 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-      - name: Set up JDK
+      - name: Set up JDK ${{ matrix.java }}
         uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java }}
           distribution: ${{ matrix.distribution }}
-      - name: Test with Maven
+      - name: Test with Maven Using Project Maven Wrapper
         run: ./mvnw clean install site -V -B -D"maven.artifact.threads=64" -D"org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
+      - name: Set up JDK 8 Integration Run (Consumer Run)
+        uses: actions/setup-java@v3
+        with:
+          java-version: 8
+          distribution: ${{ matrix.distribution }}
+      - name: Load Maven 3.2.5 Integration Run (Consumer Run) using GitHub Provided Maven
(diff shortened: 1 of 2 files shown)


hazendaz:
@mathieucarbou This is a rework of original work I had been suggesting so we can drop jdk 8 from the main build process.  If you look at the action steps on the run, it should be more clear as to what is going on.  Also the original demo I showed off was completely wrong as it was not actually running the integration tests.  It is now.  Due to just trying to better demonstrate this, I have not removed or updated checkstyle yet.  This seemed enough for now to show this.  I further dropped to maven 3.2.5 from original try of maven 3.3.9.  There is at least one maven plugin (clean plugin) that requires maven 3.2.5 or better.  Maven itself is pushing that 3.2.5 or better now in general and likely to go higher soon.  I don't know if there are really any usage stats of which maven versions but realistically anything before 3.6.x probably isn't or should not generally be used.

.github/workflows/ci.yaml
```diff
@@ -20,10 +20,19 @@ jobs:
...
-      - name: Set up JDK
+      - name: Set up JDK ${{ matrix.java }}
         uses: actions/setup-java@v3
         with:
           java-version: ${{ matrix.java }}
           distribution: ${{ matrix.distribution }}
-      - name: Test with Maven
+      - name: Test with Maven Using Project Maven Wrapper
         run: ./mvnw clean install site -V -B -D"maven.artifact.threads=64" -D"org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn"
+      - name: Set up JDK 8 Integration Run (Consumer Run)
```
mathieucarbou, line 30: My understanding is that the second task will build and not re-use the previously workspace from the first task right ?

So if we have this additionnel task that is building and running with mv n3.2.5 / java 8, why do we need to keep the jdl 8 in the matrix above ?

hazendaz:
Removal of jdk 8 would come next.  Just setting up changes here.

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

mathieucarbou (approved):

PR #619, opened 2023-08-27
[pom] Enforce byte code to java 8

hazendaz:
Main purpose here is to prevent included libraries from violating jdk requirements of project.  Currently if you set jgit to 6.6.0, it will compile without issue.  It will be caught by our usage tests on java 8, but this will further catch it right away.

Example to see, if you go to the jgit module, bump jgit to 6.6.0.202305301015-r, run a build without this, it works.  After this is added, you will get following

```
[ERROR] Rule 0: org.codehaus.mojo.extraenforcer.dependencies.EnforceBytecodeVersion failed with message:
[ERROR] Found Banned Dependency: org.eclipse.jgit:org.eclipse.jgit:jar:6.6.0.202305301015-r
[ERROR] Use 'mvn dependency:tree' to locate the source of the banned dependencies.
```

Earlier stop is more useful then waiting to make sure IT tests pass.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,8 +50,21 @@
   <properties>
     <!-- Get method parameter names via reflection -->
     <maven.compiler.parameters>true</maven.compiler.parameters>
-    <maven.compiler.source>1.8</maven.compiler.source>
-    <maven.compiler.target>1.8</maven.compiler.target>
+
+    <!-- Maven compiler options -->
+    <java.version>8</java.version>
+    <java.release.version>8</java.release.version>
+    <java.test.version>${java.version}</java.test.version>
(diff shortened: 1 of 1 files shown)


hazendaz:
Few other notes on this.

- Both source/target are legacy settings that eventually will be removed.  This broadens to use the legacy but also newer release parameter.  This later part works as we moved the build to jdk 11+.
- Split source and target options entirely so its possible if say junit goes to java 11 only (like mockito has) that we can run tests allowed at higher level.
- Three types of enforcer are caught on this protecting simple clean run, regular run, and site runs.  This generally protects all cases.

mathieucarbou (approved):

PR #629, opened 2023-09-18
[headerType] Newline on the end for slashstar_style

uyw4687:
If you see examples in java, kotlin, etc., they mostly have newline after the copyright statement. See https://github.com/mathieucarbou/license-maven-plugin/pull/186#issuecomment-686104011 

It might be because Intellij applies header file in that way and also because in this way this would be easier to divide copyright and the code.

Since customizing the headerDefinition regarding ending newline makes a quite long configuration having to reconfigure all other values, making it default might be better ƒor this plugin to be used more common.

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderType.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderType.java
@@ -43,7 +43,7 @@ public enum HeaderType {
   APOSTROPHE_STYLE("'", "' ", "'EOL", "", null, "'.*$", "'.*$", false, false, false),
   EXCLAMATION_STYLE("!", "! ", "!EOL", "", null, "!.*$", "!.*$", false, false, false),
   DOUBLEDASHES_STYLE("--", "-- ", "--EOL", "", null, "--.*$", "--.*$", false, false, false),
-  SLASHSTAR_STYLE("/*", " * ", " */", "", null, "(\\s|\\t)*/\\*.*$", ".*\\*/(\\s|\\t)*$", false, true, false),
+  SLASHSTAR_STYLE("/*", " * ", " */EOL", "", null, "(\\s|\\t)*/\\*.*$", ".*\\*/(\\s|\\t)*$", false, true, false),
   BRACESSTAR_STYLE("{*", " * ", " *}", "", null, "(\\s|\\t)*\\{\\*.*$", ".*\\*\\}(\\s|\\t)*$", false, true, false),
   SHARPSTAR_STYLE("#*", " * ", " *#", "", null, "(\\s|\\t)*#\\*.*$", ".*\\*#(\\s|\\t)*$", false, true, false),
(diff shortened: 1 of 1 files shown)


mathieucarbou:
I agree with you, but this change will break the current behaviour for existing users so this is not acceptable the way it is implemented right now.

Also, what you say might also be true for other header styles, not just java. 

So we would need to find a more global solution for that, which is backward compatible also.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderType.java
```diff
@@ -43,7 +43,7 @@ public enum HeaderType {
   APOSTROPHE_STYLE("'", "' ", "'EOL", "", null, "'.*$", "'.*$", false, false, false),
   EXCLAMATION_STYLE("!", "! ", "!EOL", "", null, "!.*$", "!.*$", false, false, false),
   DOUBLEDASHES_STYLE("--", "-- ", "--EOL", "", null, "--.*$", "--.*$", false, false, false),
-  SLASHSTAR_STYLE("/*", " * ", " */", "", null, "(\\s|\\t)*/\\*.*$", ".*\\*/(\\s|\\t)*$", false, true, false),
+  SLASHSTAR_STYLE("/*", " * ", " */EOL", "", null, "(\\s|\\t)*/\\*.*$", ".*\\*/(\\s|\\t)*$", false, true, false),
```
mathieucarbou, line 46: I agree with you, but this change will break the current behaviour for existing users so this is not acceptable the way it is implemented right now.

Also, what you say might also be true for other header styles, not just java. 

So we would need to find a more global solution for that, which is backward compatible also.

mathieucarbou (changes requested):

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/header/HeaderType.java
uyw4687: I see, it seems this issue https://github.com/mathieucarbou/license-maven-plugin/issues/78 is related.
We might need to override the number of EOLS before and after the license in (maybe) HeaderType setting. cc @minwoox

uyw4687: how about easier modification on the properties(changing only one property) instead of having to describe all other properties even though they don't need to be updated?
And application of EOL as a different issue?
since when EOL is applied, current behavior is adding that header even if there is a blank line at the beginning of the file.(though this is not a problem caused by plugin but it might be better to apply it in this way, since `onHeaderNotFound` when header is detected, header is removed. And when that header is removed, trailing blank lines are removed. It is not exactly the same case but I guess this might look natural though this is also a breaking change T.T

mathieucarbou: Your suggestions are good yes. In `HeaderType`, we can extract in a `boolean endWithEOL` state whether the style finished with EOL or not. If yes, EOL will be appended when building the header style. This new property could have a default value of false when parsed from properties or XML and this way this would be backward compatible and also won't be a bug refactoring.

Unit tests will need to be added too.

PR #636, opened 2023-09-25
[pom] Bump plexus utils to 4.0.0 and add plexus xml 3.0.0 override

hazendaz:
moving plexus forwards which includes some maven 4 in plexus xml so maven provided a bridge for plexus xml 3.0.0 to avoid maven 4 apis and thus remain compatible to maven 3.


hazendaz:
jdk 21 issue fixed on separate PR.

PR #669, opened 2023-11-07
Build updates to help reduce dependabot pull requests as well as now build with jdk 22-ea

hazendaz:


hazendaz:
ok waiting on groovy 4.0.16 for jdk 22 support which should be out soon.  No rush here.

mathieucarbou (approved):

PR #682, opened 2023-12-10
[pom] Drop fast profile as entirely unnecessary (saves 19 seconds)

hazendaz:
does not save enough time and performs a long standing hack to force execution to none.  Savings is far too little to support this.  Even if wanted, there are easier ways such as just setting skip flags that would likely produce same results but reality is this build takes 1 minute so its not really critical to skip plugins here.


mathieucarbou:
I agree;-)

mathieucarbou (approved):

PR #684, opened 2023-12-10
Use Charset throughout code (entry is still string from maven) and use NIO on file utils

hazendaz:


mathieucarbou (approved):
Nice cleanup! thanks!

PR #692, opened 2023-12-21
Minor code cleanup: ternary if statement and lambda

hazendaz:


hazendaz:
@mathieucarbou On review, you will want to hide whitespace to see this easier.

mathieucarbou (approved):

mathieucarbou:
thanks for all these recent PRs  👍

hazendaz:
@mathieucarbou Sure no problem, I'm trying to hunt down an issue where this plugin is incredibly slow and CPU spikes.  I've had little luck tracking that down as of yet but everything else is just nice to have :)

For reference in case you are wondering, with https://github.com/mybatis/mybatis-3, it takes well over a minute to run through this plugin only with git checks on dates.  That project is old and has significant git history so that is likely most of the reason but I suspect some leaks in the code potentially cause the problem.  I do have a highly powerful machine and the behavior is very consistent on mine and has been enough that on most github projects as well as work projects, where possible we turn this plugin off due to this.  I believe that is why there are general asks to get smarter about handing dates in general ;)

mathieucarbou:
Pretty sure this is the issue: going through the commit range for the dates, and for each file, is really expensive.
The code is also not improved a lot and quite sequential: ideally, it would need to be refactored and use maybe a sort of work stealing queue with N threads, N being the number of cores of the machine, and be able to handle in one time a batch of files. This is a big rework though I think because all the current processing is per file.

mathieucarbou:
Note: maybe create an issue for that, flagged as `todo` and `bug` ?

Issue #691, opened 2023-12-21
Release parent-pom to fix double 'source' runs


hazendaz:
See https://github.com/mathieucarbou/parent-pom/pull/8

I've updated all the plugins, fixed the issue with source there as we are double running (this will allow us to upgrade), and added some protections on javadocs from a couple aspects.

If possible, would like to get this out sooner rather than later so we can fix up the license plugin build :)

PR #693, opened 2023-12-21
Drop parent pom, fix up some build issues, add new discovery of .checkstyle now being produced in recent Eclipse with personal data

hazendaz:
Fixes https://github.com/mathieucarbou/license-maven-plugin/issues/691

There is a lot here, normally I would have ended up splitting things up to smaller pulls but given having to run side by side comparisons with existing build in favor of dropping the parent-pom from the mix, I felt it was easier to just keep all findings together and some of that journey to get here.  You may want to also run side by side comparison to confirm.  

## Differences in the build ##
- Before, as known, source was running in both forked and non forked mode.  It now runs exactly as its supposed to and can be on latest version now.
- Maven plugin descriptor and help mojo's were running twice as the naming in parent-pom id and this project were different

## Things left behind in parent-pom ##
- Did not pull forwards any of the OSGI stuff, none of it ran against this project
- Did not pull forwards a groovydoc plugin that isn't used here either ( was in release to attach but presume it just skipped )
- Dropped some settings that were not needed, added others

## Build Review ##
Side by side of the builds was reviewed.  The internal usage of license plugin check now runs before checkstyle and the second run of enforcer is moved after those.  Only warnings left are 'our' deprecation(s) and usage of deprecated maven features.

## Checkstyle ##
Newly discovered with Eclipse 2023-12 checkstyle integration is that its creating .checkstyle files.  I don't know if those files are new but had not seen that before and they started showing up in last week or so which I had been ignoring.  I've added a temporary ignore in our build for internal check and added this to general ignore defaults.  It does have disk identifying data in it so ignoring it makes most sense as unlikely checked in items.

## Follow-ups ##
As a follow up on javadoc issue, fixed the last remaining alerted warning.

As a fix to internal license, made sure the excludes is inherited, this was noticed when .checkstyle was added in parent but then ignored in the plugin module.

## Junit 5 ##
I also noticed junit 5 vintage usage and tried to remove that but its still needed somehow by maven plugin even though we updated to their junit 5 copy.  They do have it noted as optional and its failing with junit 3 classes being missing which we are not using and I could only find in one test case of mavens so not exactly sure where they are pulling that as the stack trace seemed to indicate that project causes it.  When adding vintage back though, I did some work on hamcrest as the 'core' module is empty and long deprecated to use the main one (all code was merged into single module).  So that is cleaned up so we don't have both hamcrest and hamcrest-core floating around.  I did note where exactly this vintage usage is needed.  If that is fixed in some future release of maven testing harness which I'd suspect would be, we would still need hamcrest as we are using it at the moment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,57 +112,6 @@
       </properties>
     </developer>
   </developers>
-  <contributors>
-    <contributor>
-      <name>Peter Palaga</name>
-      <email>ppalaga@redhat.com</email>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>masakimu</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>Matthieu Brouillard</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>Royce Remer</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>Juergen Hoeller</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>Rob Harrop</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>Cedric Pronzato</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-    <contributor>
-      <name>Michael J. Simons</name>
-      <roles>
-        <role>Developer</role>
-      </roles>
-    </contributor>
-  </contributors>
 
   <licenses>
     <license>
--- a/pom.xml
+++ b/pom.xml
@@ -414,7 +414,6 @@
           <version>${maven-plugin.version}</version>
           <configuration>
             <detail>true</detail>
-            <skipErrorNoDescriptorsFound>true</skipErrorNoDescriptorsFound>
           </configuration>
         </plugin>
         <plugin>
@@ -623,24 +622,6 @@ limitations under the License.
               <goal>enforce</goal>
             </goals>
             <phase>pre-site</phase>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-plugin-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>generate-helpmojo</id>
-            <goals>
-               <goal>helpmojo</goal>
-            </goals>
-          </execution>
-          <execution>
-            <id>generate-descriptor</id>
-            <goals>
-              <goal>descriptor</goal>
-            </goals>
           </execution>
         </executions>
       </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -31,12 +31,12 @@
 
   <issueManagement>
     <system>GitHub</system>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -169,6 +166,14 @@
...
+    <license>
+      <name>Apache License 2.0</name>
+      <url>http://www.apache.org/licenses/LICENSE-2.0.html</url>
+      <distribution>repo</distribution>
+    </license>
+  </licenses>
+
   <scm>
     <connection>scm:git:https://github.com/mathieucarbou/${mycila.github.name}.git</connection>
     <developerConnection>scm:git:git@github.com:mycila/${mycila.github.name}.git</developerConnection>
```
mathieucarbou, line 179: placeholders could be replaced here

pom.xml
```diff
@@ -261,6 +266,15 @@
...
 
   <distributionManagement>
+    <repository>
+      <id>mycila-oss-staging</id>
+      <url>https://oss.sonatype.org/service/local/staging/deploy/maven2/</url>
+    </repository>
+    <snapshotRepository>
+      <id>mycila-oss-snapshots</id>
+      <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
+      <uniqueVersion>false</uniqueVersion>
```
mathieucarbou, line 276: pretty sure we can drop the snapshot repo

pom.xml
```diff
@@ -169,6 +166,14 @@
     </contributor>
   </contributors>
```
mathieucarbou, line 167: You could have your id here (and eventually some other contributors too), or, more more simplicity, we drop this section and only maintain the one in the website ?

pom.xml
```diff
@@ -628,6 +655,34 @@ limitations under the License.
...
+          <plugin>
+            <groupId>com.mycila</groupId>
+            <artifactId>license-maven-plugin</artifactId>
+            <configuration>
+              <skip>true</skip>
+            </configuration>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-gpg-plugin</artifactId>
```
mathieucarbou, line 672: About that, could you please add `<useAgent>true</useAgent>` in this plugin config ? I always run into the same failure and have to manually trigger the upload command again because it is missing ;-)

mathieucarbou (changes requested):

pom.xml
hazendaz: @mathieucarbou Given we use git, it makes sense to drop it entirely as that data can come from git.

pom.xml
hazendaz: @mathieucarbou Its in dependency management set to 'true' on line 407.

pom.xml
hazendaz: ok will take care of cleaning up some of that, I was thinking the same when doing this :)

pom.xml
hazendaz: I'm leaving developer section since it has metadata to find us.  The contributor data as a whole did not other than names and as noted that can be found in git.

hazendaz: actually one did have an email but still probably doesn't matter much.

hazendaz:
@mathieucarbou Items noted should be good now, master was merge conflicted due to groovy PR so its merged in here, did that from GH.

mathieucarbou (approved):

mathieucarbou:
> @mathieucarbou Items noted should be good now, master was merge conflicted due to groovy PR so its merged in here, did that from GH.

the problem is that I cannot force merge with conflicts...

mathieucarbou:
Ok I had to introduce a merge commit.

PR #752, opened 2024-04-27
Update maven wrapper to 3.3.1 and adjustment in renovate tracking file

hazendaz:
Renovate file for my fork usage and otherwise entirely ignored.  Maven wrapper updated to 3.3.1.

### DIFF
--- a/.mvn/wrapper/MavenWrapperDownloader.java
+++ b/.mvn/wrapper/MavenWrapperDownloader.java
@@ -21,77 +21,72 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.Authenticator;
 import java.net.PasswordAuthentication;
+import java.net.URI;
 import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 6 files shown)


mathieucarbou (approved):

mathieucarbou:
2 builds are failing

hazendaz:
> 2 builds are failing

Looks like macos no longer supports java 8.  Unrelated to this PR.

hazendaz:
> > 2 builds are failing
> 
> Looks like macos no longer supports java 8. Unrelated to this PR.

See https://github.com/actions/setup-java/issues/625, m1 for temurin does not support java 8.  Found a patch off that thread that does work that I can raise separately.  As an aside, we need to drop usage of maven before 3.6.3 per maven policies and all maven core plugins are starting to now require 3.6.3.  We have this running java 8 with maven 3.2.5 now so I'll add missing entry in the maven plugin to make sure it denotes 3.6.3 required and raise that up.  May bring those items together to address macos situation faced here.

mathieucarbou (approved):

Issue #859, opened 2024-11-23
‎Folders are not properly added to the folder exclusion list when scanning files on Windows.


EvanTich:
**Version affected**

Running on the latest version (4.6) on Windows 10.

**Describe the bug**

Building the exclusions in `Selection::buildExclusions` adds all excludes to the list without regard for the system's path-separator. The `DirectoryScanner` used in `Selection::scanIfneeded` automatically converts file separators to the proper ones, including the `name` used in the `ScanConductor`, so folders that should be excluded are not.

This is largely a benign bug that only shows its fangs when large, auto-generated folders like `node_modules` or `.vite` are created.

**How to Reproduce**

1. Open any Maven project that uses this plugin.
2. Run `license:format` with debug enabled. The bug should also be able to be seen on `license:check` and `license:remove`.
3. Find line stating `Starting to visit {basedir}, excluding directories: []`. Observe default folders and others are not properly added to folder excludes.
4. Add an exclude like `**\node_modules\**` to the plugin settings/parameters.
5. Re-run the command with debug, and see the directory is excluded correctly.

Hopefully, that should be enough to reproduce the bug...

hazendaz:
Do you believe it is reading those folders and files?  Our exclusion list is pretty large and it doesn't complain about inability to read things like jpg for example.  Is it just that its not showing you the default list correctly and that would help for debugging purposes?  Since you note its benign, I presume that is the case but wanted to clarify.  Of course it should show the list correctly if discovered as that would be important for not only debugging but understanding how the plugin works.

EvanTich:
The problem is that it reads every file name, not the contents, in folders that should be excluded. I say that it is benign **until** you get huge folders like `node_modules` in the mix which can have hundreds or thousands of files. Execution slows to a halt until each file name in the folder is collected recursively, with all of them being ignored anyway down the line.

The `ScanConductor` implementation inside `Selection::scanIfneeded` takes in file names with the appropriate path-separator and checks that name with the folder excludes, see [this line in the `ScanConductor`](https://github.com/mathieucarbou/license-maven-plugin/blob/master/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/Selection.java#L85). Unfortunately, the exclusion file names are improperly formatted with `/` path-separators, so that method never tells the `DirectoryScanner` to not recurse through that folder. I haven't had the time to properly debug the plugin, but that's what I think is happening.

hazendaz:
[...]

PR #870, opened 2024-12-26
WIP Rewrite selection code

hazendaz:
Code is invalid, confirmed on mybatis project that licensing incredibly speeds up by fixing this and still does what it should.  Tests currently fail as they are hard-coded to LF not OS dependent.  Need to fix that but pushing up to get a review.

### DIFF
--- a/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/Selection.java
+++ b/license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/Selection.java
@@ -100,19 +100,24 @@ public final class Selection {
     }
   }
 
-  private String[] findFolderExcludes() { // less we keep, less overhead we get so we only use user excludes there
-    final List<String> excludes = new ArrayList<>(excluded.length / 2 /*estimate*/);
-    for (final String exclude : (userExcluded != null ? userExcluded : excluded)) {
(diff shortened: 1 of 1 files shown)


hazendaz:
See #859 that found the issue.  Tests are currently hard-code to LF so they fail on windows.  I presume will work here on *nix but haven't confirmed at time of writing.  I used the build by skipping tests on mybatis which was taking minutes to process, now seems faster but need to do more testing to confirm.

@mathieucarbou Can you look this over to see if I'm on the right path here?

hazendaz:
ok ignore my mybatis test results.  Variance is a windows thing so it doesn't actually improve with this for that project.  Seems rather something on my machine changed so I was getting better results but retesting its master I get seconds of same result so this has no improvement there. Will it improve projects with node_modules, probably but its a read so I doubt it changes much.  I still think original logic is indeed incorrect but not sure what was in mind back in 2020 and regardless issue I thought this solved I think is longer standing than 2020.

hazendaz:
ok GHA confirmed.  Tests need simply updated to work cross platform.  Provided this is on right track, will address the tests afterwards.  *nix worked fine per GHA.

hazendaz:
will circle back on this after other work.  For now going to close.

Issue #536, opened 2023-04-18
Command Line flag to alter the list of included files to check


mathieu:
**Describe the feature**

Using the plugin to check only a set of files (via precommit). It works fine with the legacy way of declaring include files. 

```
mvn license:check -Dlicense.includes=src/main/myfile.java
```

But if I move my configuration inside  the proposed `<LicenseSets><LicenseSet></LicenseSet></LicenseSets>` section 

using the `LicenseSet.includes` proposed alternative doesn't work and all files are checked.

```
mvn license:check -DLicenseSet.includes=src/main/myfile.java
```

**Any alternative you've considered ?**

Keep the legacy configuration but it has deprecation warnings

```
[INFO] --- license:4.2:check (default-cli) @ runtime-parent ---
[WARNING] Parameter 'legacyConfigExcludes' (user property 'license.excludes') is deprecated: use LicenseSet.excludes
[WARNING] Parameter 'legacyConfigIncludes' (user property 'license.includes') is deprecated: use LicenseSet.includes
[WARNING] Parameter 'legacyConfigInlineHeader' (user property 'license.inlineHeader') is deprecated: use LicenseSet.inlineHeader
```

hazendaz:
have you tried the following?

```
mvn license:check -DlicenseSets.licenseSet.includes=src/main/myfile.java
```

and if done from windows, quoted instead...

```
mvn license:check -D"licenseSets.licenseSet.includes=src/main/myfile.java"
```

mathieu:
An alternative proposal could be to have a top-level variable that would list the files to consider instead of the whole content of the current folder...

mathieu:
I will try to make a PR proposing to have a top-level variable that would list the files to consider instead of the whole content of the current folder...

PR #876, opened 2025-01-07
feat: Add a property to limit plugin to a list of files

mathieu:
The idea is to be able to use this plugin from a pre-commit hook (only run on a set of files)

usage: mvn license:check -Dlicense.filesToCheck="src/main/org/acme/MyClass.java,src/main/org/acme/MyEnum.java"

ref #536

### DIFF
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/invoker.properties
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/invoker.properties
@@ -1,0 +1,1 @@
+invoker.goals=license:format
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/mock-license.txt
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/mock-license.txt
@@ -1,0 +1,2 @@
+This is a
+mock license
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/pom.xml
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/pom.xml
@@ -1,0 +1,34 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.mycila.license-maven-plugin.it</groupId>
+  <artifactId>legacy-config</artifactId>
+  <version>1.0.0</version>
+
+  <name>Check Legacy Configuration with checkFiles</name>
+  <description>Integration Test for checking legacy configuration with checkFiles</description>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>com.mycila</groupId>
+        <artifactId>license-maven-plugin</artifactId>
+        <version>@project.version@</version>
+        <configuration>
+          <header>mock-license.txt</header>
+          <excludes>
+            <exclude>invoker.properties</exclude>
+            <exclude>pom.xml</exclude>
+            <exclude>*.groovy</exclude>
+            <exclude>**/*.bak</exclude>
+            <exclude>*.log</exclude>
+          </excludes>
+          <filesToCheck>src/main/java/com/mycilla/it/Unformatted2.java</filesToCheck>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/setup.groovy
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/setup.groovy
@@ -1,0 +1,23 @@
+import java.nio.file.Files
+import java.nio.file.Path
+
+final Path base = basedir.toPath()
+
+if (!Files.exists(base) || !Files.isDirectory(base)) {
+  System.err.println("base directory is missing.")
+  return false
+}
+
+final List<String> ALL_FILES = Arrays.asList("Unformatted1.java", "Unformatted2.java")
+
+for (final String filename : ALL_FILES) {
+  final Path unformattedJavaFile = base.resolve("src/main/java/com/mycilla/it/" + filename)
+  if (!Files.exists(unformattedJavaFile)) {
+    System.err.println(filename + " file is missing.")
+    return false
+  }
+
+  Files.copy(unformattedJavaFile, unformattedJavaFile.resolveSibling(filename + ".bak"))
+}
+
+return true;
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/src/main/java/com/mycilla/it/Unformatted1.java
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/src/main/java/com/mycilla/it/Unformatted1.java
@@ -1,0 +1,4 @@
+package com.mycila.it;
+
+public class Unformatted1 {
+}
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/src/main/java/com/mycilla/it/Unformatted2.java
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/src/main/java/com/mycilla/it/Unformatted2.java
@@ -1,0 +1,4 @@
+package com.mycila.it;
+
+public class Unformatted2 {
+}
--- a/license-maven-plugin/src/it/legacy-config-checkFiles/verify.groovy
+++ b/license-maven-plugin/src/it/legacy-config-checkFiles/verify.groovy
@@ -1,0 +1,61 @@
+import java.nio.file.Files
+import java.nio.file.Path
+
+import static org.junit.jupiter.api.Assertions.assertEquals
+
+
+final Path base = basedir.toPath()
+
+if (!Files.exists(base) || !Files.isDirectory(base)) {
+  System.err.println("base directory is missing.")
+  return false
+}
+
+final Path license = base.resolve("mock-license.txt")
+if (!Files.exists(license)) {
+  System.err.println("license file is missing.")
+  return false
+}
+
+final Path unformattedJavaFile1 = base.resolve("src/main/java/com/mycilla/it/Unformatted1.java.bak")
+if (!Files.exists(unformattedJavaFile1)) {
+  System.err.println(unformattedJavaFile1.getFileName() + " file is missing (should have been created by setup.groovy).")
+  return false
+}
+
+final Path unformattedJavaFile2 = base.resolve("src/main/java/com/mycilla/it/Unformatted2.java.bak")
+if (!Files.exists(unformattedJavaFile2)) {
+  System.err.println(unformattedJavaFile2.getFileName() + " file is missing (should have been created by setup.groovy).")
+  return false
+}
+
+final Path formattedJavaFile1 = base.resolve("src/main/java/com/mycilla/it/Unformatted1.java")
+if (!Files.exists(formattedJavaFile1)) {
+  System.err.println(formattedJavaFile1.getFileName() + " file is missing.")
+  return false
+}
+
+
+final Path formattedJavaFile2 = base.resolve("src/main/java/com/mycilla/it/Unformatted2.java")
+if (!Files.exists(formattedJavaFile2)) {
+  System.err.println(formattedJavaFile2.getFileName() + " file is missing.")
+  return false
+}
+
+assertEquals(new String(Files.readAllBytes(formattedJavaFile1)), new String(Files.readAllBytes(unformattedJavaFile1)))
+
+final StringBuilder expected = new StringBuilder();
+expected.append("/*\n");
+license.withReader { reader ->
+  while ((line = reader.readLine()) != null) {
+    expected.append(" * ").append(line).append('\n')
+  }
+}
+expected.append(" */\n")
+expected.append(new String(Files.readAllBytes(unformattedJavaFile2)))
+
+final String actual = new String(Files.readAllBytes(formattedJavaFile2))
+
+assertEquals(expected.toString(), actual)
+
+return true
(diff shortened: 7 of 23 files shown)


mathieucarbou:
Thanks! We'll look!

mathieucarbou:
From a quick reading I see that the property impacts all goals so I would rename the property to something like: `globalFileFilter` or an equivalent name not bound to any goal but more to the fact that this actually replaces the file scan.

Also, after reading #536, it feels like this is much more a big fix ? Right @hazendaz ?

mathieu:
> From a quick reading I see that the property impacts all goals so I would rename the property to something like: `globalFileFilter` or an equivalent name not bound to any goal but more to the fact that this actually replaces the file scan.

As you understood, the intent is to replace the file scan by giving a list of files to run the plugin goals on. The property should reflect this, it is not a filter per se.

> Also, after reading #536, it feels like this is much more a big fix ? Right @hazendaz ?
> 
> Passing properties as proposed here [#536 (comment)](https://github.com/mathieucarbou/license-maven-plugin/issues/536#issuecomment-1519261086) should work... It is also documented in the plugin javadoc (from memory)

The proposed solutions did not work for me

mathieucarbou:
> The property should reflect this, it is not a filter per se.

It's neither a "check" (ref: `filesToCheck`)... 

Like said: _or an equivalent name not bound to any goal but more to the fact that this actually replaces the file scan._

mathieu:
> That is my real concerne here.... Before merging this PR, I would like to understand why this is not working and whether we should make it work... These properties being documented, I would expect them to work when passed as parameter. They were also working before the move to a list structure.

I'll try to create a test to reproduce this but this is the debug redacted output of a local license:check configuration

```log
[DEBUG] Loading mojo com.mycila:license-maven-plugin:4.6:check from plugin realm ClassRealm[plugin>com.mycila:license-maven-plugin:4.6, parent: jdk.internal.loader.ClassLoaders$AppClassLoader@76ddd772]
[DEBUG] Configuring mojo execution 'com.mycila:license-maven-plugin:4.6:check:default-cli' with basic configurator -->
[DEBUG]   (f) aggregate = true
[DEBUG]   (f) concurrencyFactor = 1.5
[DEBUG]   (f) defaultBasedir = [...]
[DEBUG]   (f) defaultProperties = {inceptionYear=2018, license.git.maxCommitsLookup=1, license.warnIfShallow=false}
[DEBUG]   (f) defaultUseDefaultExcludes = true
[DEBUG]   (f) dependencyEnforce = false
[DEBUG]   (f) dependencyExceptionMessage = Some licenses were denied by policy:
[DEBUG]   (f) dependencyPolicies = []
[DEBUG]   (f) dependencyScopes = [runtime]
[DEBUG]   (f) dryRun = false
[DEBUG]   (f) encoding = UTF-8
[DEBUG]   (f) errorMessage = Some files do not have the expected license header. Run license:format to update them.
[DEBUG]   (f) failIfMissing = true
[DEBUG]   (f) failIfUnknown = true
[DEBUG]   (f) legacyConfigExcludes = []
[DEBUG]   (f) legacyConfigIncludes = []
[DEBUG]   (f) inlineHeader = [...]
[DEBUG]   (f) includes = [**/src/**, **/cli-gen/**, **/i18n/**, *.sh]
[DEBUG]   (f) excludes = [**/src/main/resources/**/*.key, **/src/main/resources/**/*.crt, **/src/test/resources/**, **/src/main/resources/selectors/**, **/src/main/resources/*.avsc, **/src/main/resources/*.sql]
[DEBUG]   (f) licenseSets = [com.mycila.maven.plugin.license.LicenseSet@855f5953]
[DEBUG]   (f) mapping = {bats=SCRIPT_STYLE, conf=DOUBLESLASH_STYLE, jvm.options=SCRIPT_STYLE, less=DOUBLESLASH_STYLE, ltpa.keys=SCRIPT_STYLE, proto=DOUBLESLASH_STYLE, scss=SLASHSTAR_STYLE, ts=SLASHSTAR_STYLE, tsx=SLASHSTAR_STYLE}
[DEBUG]   (f) nThreads = 0
[DEBUG]   (f) prohibitLegacyUse = false
[DEBUG]   (f) project = MavenProject: [...]/LATEST-SNAPSHOT @ [...]/pom.xml
[DEBUG]   (f) quiet = false
[DEBUG]   (f) reportLocation = [...]/target/site/license-plugin-report.xml
[DEBUG]   (f) reportSkipped = false
[DEBUG]   (f) session = org.apache.maven.execution.MavenSession@a1e84b7a
[DEBUG]   (f) settings = org.apache.maven.execution.SettingsAdapter@7315678a
[DEBUG]   (f) skip = false
[DEBUG]   (f) skipExistingHeaders = true
[DEBUG]   (f) strictCheck = true
[DEBUG]   (f) useDefaultMapping = true
[DEBUG]   (f) warnIfShallow = true
[DEBUG] -- end configuration --
```

This line is triggering: `[DEBUG]   (f) licenseSets = [com.mycila.maven.plugin.license.LicenseSet@855f5953]`

The call was made with `-DlicenseSets.licenseSet.includes=pom.xml`

mathieucarbou:
Right, there is no toString() representation sadly - one should be added, which displays the object content.

mathieu:
@mathieucarbou, I've created a new PR with a failing IT test: #878

mathieu:
Proposition for the name of the property to set the files to replace the scan of the `baseDir`: 

1. `files`, simple and straightforward, indicating a collection of files.
4. `filePaths`, indicates the list holds the file paths of the files
5. `fileList`,  Clearly communicates that this is a list of files
6. `filesToCheck`, _Already discarded because it is not a check_
7. `globalFileFilter`, Misleading as this property holds list of files not a filter
8. other propositions

@mathieucarbou and @hazendaz, do you have any preferences ?

hazendaz:
@mathieu I haven't looked at this yet but seems like something we are looking for in mybatis.  We are looking to add the git commit hook plugin for any formatting tools because its too frequent that folks use the IDE to make changes, test them, etc but then never run the full build and miss formatting only to make junk commits later.  I've done extensive profiling on more than 100 plugins and this by far when messing with the copy right dates is the slowest plugin that exists.  That said, this seems interesting to get working so it can work off a diff only as no reason to check all files constantly.  Another PR currently out to better handle the slow git commands seems like a double punch to make this plugin go from last to somewhere reasonable.  I'm going to try to review all you have up including the alternative this weekend but may take a little longer to fully understand all the changes.

I think 3 in your list seems appropriate in case I'm thinking but before doing any more, give us some time to review fully so we better understand and test this against mybatis where we already have an open PR to add in the git commit hook plugin (without license right now).  Currently doing that there ends up taking 8 minutes on the contributors machine, mine takes far less but I have a bulked up machine that is serious overkill for this.  At work though, I'm on a dev ops team supporting some 2k+ repos and due to performance on this plugin, we have not really promoted its usage but want to.  I think if all these current pull requests work out, it will allow us to further promote that.  It may not be a silver bullet but hopefully its enough to make it viable for using a hook to address as code is committed without killing developer experience.

mathieucarbou:
@mathieu @hazendaz : I was thinking about this feature this morning. 

Like @mathieu said, the goal is to globally pre-filter the set of files that the plugin will use to work with (pre-commit, etc), which is current done by replacing the scanner.

I think we can do better.

The main idea is to reduce the selection on which the plugin will work, but the way to define this selection, we already have something we could re-use in terms of plugin: the concept of includes/excludes with ant-like patterns, which is even more flexible than defining specific file paths.

This concept is also hidden in the name proposals here: https://github.com/mathieucarbou/license-maven-plugin/pull/876#issuecomment-2582311857:

In maven, we already have the concepts in other plugins (i.e. fileSets and source for assembly - https://maven.apache.org/plugins/maven-assembly-plugin/assembly.html). So why not re-use this concept and introduce something like this.

It cannot be  a global `<includes>` or `<excludes>` since they already exist and have been deprecated.

But it could be something really conveying the fact that:

1) this is filtering globally the sources / workspace
2) this is using nat patterns (which exclude `file` proposals because they are not files but patters)

If set, then, we can use the scanner as usual to list the files we should work with, but this list, which forms a virtual workspace, then has to be filtered again using the  include/exclude of each license set.

Some thing like (adapting an example from the project and using `<workspace>`)

```xml

  <build>
    <plugins>
      <plugin>
        <groupId>com.mycila</groupId>
        <version>@project.version@</version>
        <configuration>

          <workspace>
            <includes>
                <include>specific/file.java</include>
                <include>all/in/this/folder/.java</include>
            <includes>
            <excludes>
                <exclude>all/in/this/folder/DoNotTouch.java</exclude>
            <excludes>
          <workspace>
          
          </fileSets>
          <licenseSets>
            <licenseSet>
              <header>mock-license-1.txt</header>
              <excludes>
                <exclude>invoker.properties</exclude>
                <exclude>pom.xml</exclude>
                <exclude>*.groovy</exclude>
                <exclude>**/*.bak</exclude>
                <exclude>*.log</exclude>
                <exclude>mock-license-*</exclude>
                <exclude>**/Unformatted2.java</exclude>
                <exclude>**/Unformatted3.java</exclude>
              </excludes>
            </licenseSet>
            <licenseSet>
              <header>mock-license-2.txt</header>
              <excludes>
                <exclude>invoker.properties</exclude>
                <exclude>pom.xml</exclude>
                <exclude>*.groovy</exclude>
                <exclude>**/*.bak</exclude>
                <exclude>*.log</exclude>
                <exclude>mock-license-*</exclude>
                <exclude>**/Unformatted1.java</exclude>
                <exclude>**/Unformatted3.java</exclude>
              </excludes>
            </licenseSet>
            <licenseSet>
              <header>mock-license-3.txt</header>
              <includes>
                <include>**/Unformatted3.java</include>
              </includes>
            </licenseSet>
          </licenseSets>
        </configuration>
      </plugin>
    </plugins>
  </build>
```

mathieu:
So using this proposal from pre-commit would mean something like this ?

mvn license:check -Dlicense.workspace.includes=src/file1.txt,src/file2.txt

mathieucarbou:
> So using this proposal from pre-commit would mean something like this ?
> 
> mvn license:check -Dlicense.workspace.includes=src/file1.txt,src/file2.txt

Yes! it would be easier to use it like that and also with patterns.

Also, if you look at https://github.com/mathieucarbou/license-maven-plugin/pull/878, it works actually, but more a pain to setup in a commit hook.

mathieucarbou:
@hazendaz : what do you think ? Should we release with the latest pr merged or we wait to have the work for this one done ? 
I like the idea of constraining the work to a more limited workspace.

But on the other end, as shown, there is currently a way to use a commit hook to only format the committed files, although not efficient.

hazendaz:
> @hazendaz : what do you think ? Should we release with the latest pr merged or we wait to have the work for this one done ? I like the idea of constraining the work to a more limited workspace.
> 
> But on the other end, as shown, there is currently a way to use a commit hook to only format the committed files, although not efficient.

I think we should wait for this.  I'm also refocusing over here to do some other cleanup.  Since we are going into our first version 5 release on java 11 now its already going to show as a breaking change in terms of the 4.x stream so multiple bigger feature improvements seems appropriate in this particular case even if we are mostly sticking as close to underlying current api as possible.

mathieucarbou:
> Since we are going into our first version 5 release on java 11

You're right! Forgot about that!

mathieucarbou:
@mathieu : are you willing to update your PR to try the proposal about with the workspace idea ?

Also, `defaultBasedir` should be moved into this new workspace object and the existing one be deprecated and its defaukt value removed.

```xml
          <workspace>
            <baseDir></baseDir>
            <includes>
                <include>specific/file.java</include>
                <include>all/in/this/folder/.java</include>
            <includes>
            <excludes>
                <exclude>all/in/this/folder/DoNotTouch.java</exclude>
            <excludes>
          <workspace>
```

mathieu:
@mathieucarbou , I'll need some more context on your vision of the `<workspace>`.  What does it do compared to LicenceSet's `<includes>` and `<excludes>`  ?

mathieucarbou:
> @mathieucarbou , I'll need some more context on your vision of the `<workspace>`. What does it do compared to LicenceSet's `<includes>` and `<excludes>` ?

`workspace` would define the global workspace to work one. 

- workspace.baseDir == project root directory
- workspace.includes defaulted to null
- workspace.excludes defaulted to null

This concept exists already but is implicit: the workspace.baseDir is currently the workspace.baseDir, we do not constrain to any include (the whole workspace is visible) and we do not constrain to some excludes (nothing is excluded by default).

The idea is not make it explicitly configurable.

`LicenceSet` are used to "split" the workspace into several groups of file and assigning them a license. So they inherit the workspace baseDir, includes, excludes. licensSet.baseDir can be overridden but must be a sub directory of the workspace (@hazendaz : I don't know if we make sure of that somewhere, but with the workspace concept we will be)

As I see things for includes/excludes:

- if workspace.includes is defined in POM, then it will be added to all licenseSet.includes
- if workspace.excludes is defined in POM, then it will be added to all licenseSet.excludes
- if workspace.baseDir is defined in POM, then it will be used to set licenseSet.baseDir (if not set), and if licenseSet.baseDir is set, we have to check that it is same or a subfolder of workspace.baseDir

In the case of a commit hook, then the only property to add would be:

`-D license.workspace.includes=<list of file paths>`

This behind will be transferred to licenseset.includes and will have at the end the same effect of my comment in this PR: https://github.com/mathieucarbou/license-maven-plugin/pull/878#issuecomment-2585331008

Also, this "design" does not change the scanner logic, which is still taking place but based on a refinement of the baseDir / includes / excludes.

mathieu:
In other words, we do not bypass the `DirectoryScanner` anymore ?

mathieucarbou:
> In other words, we do not bypass the `DirectoryScanner` anymore ?

No because if you bypass it, you cannot support ant patterns.

And it is also less code to not bypass it and reuse the include / exclude mechanism.

mathieu:
I'll close this PR and open an issue (#887) for the workspace feature. I'll open a new PR with updates for it

PR #877, opened 2025-01-07
lazy evaluation for git provided properties

### LINKED ISSUE #874: Speed up resolution of git provided properties for header content
**Describe the feature**

Using header properties provided from git, like `license.git.copyrightLastYear`, can be slow, 
even used together with 
```
<licenseSets>
  <licenseSet>
    <properties>
       <license.git.maxCommitsLookup>1</license.git.maxCommitsLookup>
    </properties>
...
```

That's because the two Property providers in `com.mycila.maven.plugin.license.git` use seperate JGit rev walks for (nearly) every property their provide, and all properties are eagerly provided.

For projects not using all provided properties to build header text this will result in unnecessary rev walks.

**Any alternative you've considered ?**

No.

frawa:
Resolves #874.

### DIFF
--- a/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightRangeProvider.java
+++ b/license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightRangeProvider.java
@@ -15,17 +15,17 @@
  */
 package com.mycila.maven.plugin.license.git;
 
-import java.io.File;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicReference;
-import java.util.function.Supplier;
 
 import org.eclipse.jgit.api.errors.GitAPIException;
 
 import com.mycila.maven.plugin.license.AbstractLicenseMojo;
 import com.mycila.maven.plugin.license.PropertiesProvider;
 import com.mycila.maven.plugin.license.document.Document;
+import com.mycila.maven.plugin.license.util.LazyMap;
+import com.mycila.maven.plugin.license.util.Fn;
 
 /**
  * An implementation of {@link PropertiesProvider} that adds {@value #COPYRIGHT_LAST_YEAR_KEY} and
@@ -60,8 +60,7 @@ public class CopyrightRangeProvider implements PropertiesProvider {
   }
 
   /**
-   * Returns an unmodifiable map containing the following entries, whose values
-   * are set based on inspecting git history.
+   * Returns an unmodifiable map containing the following entries, whose values are set based on inspecting git history.
    *
    * <ul>
    * <li>{@value #COPYRIGHT_LAST_YEAR_KEY} key stores the year from the committer date of the last git commit that has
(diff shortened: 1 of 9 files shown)


mathieucarbou:
Hello @frawa,
Thanks for starting such efforts.
Please make sure to use the code style present in the classes you modify and do not reformat ;-)
Happy new year!

mathieucarbou:
... also do not forget to squash your commits ;-)

hazendaz:
> ... also do not forget to squash your commits ;-)

will make sure to squash on merge.

.config/style.intellij-idea.eclipse.xml
mathieucarbou, line 1: please do not forget to remove this file from the PR (eventually adding an entry in .ignore file)

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightRangeProvider.java
```diff
@@ -76,50 +78,60 @@ public void close() {
...
+          } else if (COPYRIGHT_CREATION_YEAR_KEY.equals(key)) {
+            return gitLookup.getYearOfCreation(document.getFile());
+          }
+          return null;
+        } catch (IOException | GitAPIException e) {
+          throw new RuntimeException(
+              "CopyrightRangeProvider error on file: " + document.getFile().getAbsolutePath() + ": " + e.getMessage(), e);
+        }
+      });
+    };
```
mathieucarbou, line 109: Even if I understand the memoizing effect of it, I think it can be written in a better form for code readability and future maintainance by introduce a static method:

```
    public static <T, R> Function<T, R> memoize(Function<T, R> function) {
        Map<T, R> cache = new HashMap<>();
        return input -> cache.computeIfAbsent(input, function);
    }
```

and then crating only 2 functions for the 2 cases.

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
```diff
@@ -86,35 +86,15 @@ public enum DateSource {
    * @return      the git lookup
    */
   public static GitLookup create(File file, Map<String, String> props) {
-    final GitLookup.DateSource dateSource = Optional.ofNullable(props.get(COPYRIGHT_LAST_YEAR_SOURCE_KEY))
-        .map(String::trim)
-        .map(String::toUpperCase)
-        .map(GitLookup.DateSource::valueOf)
-        .orElse(GitLookup.DateSource.AUTHOR);
+    final GitLookup.DateSource dateSource = Optional.ofNullable(props.get(COPYRIGHT_LAST_YEAR_SOURCE_KEY)).map(String::trim).map(String::toUpperCase).map(GitLookup.DateSource::valueOf).orElse(GitLookup.DateSource.AUTHOR);
```
mathieucarbou, line 89: can you please revert to the original format of the file and keep the existing code style of a source code ? this is to keep the diffs focused on the PR goal - no code reformat in the PRs. ;-) Thanks!

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -669,7 +671,7 @@ private void executeForLicenseSet(final LicenseSet licenseSet, final Callback ca
               .map(Objects::toString).collect(Collectors.joining("\n - ")));
         }
 
-        return perDoc;
+        return new AdjustedMap<>(perDoc, perDocAdjustments);
```
mathieucarbou, line 674: I like this hack, but it could be improved. The `perDoc` map could be this adjusted map (called differently - could be named `ForwardingMap`), where keys would be first fetched from the suppliers, otherwise from `file.name` otherwise from globalProperties maps.

there is

mathieucarbou (commented):
A few improvements and cleanup to do.
Thanks! This is a nice speed improvement!

.config/style.intellij-idea.eclipse.xml
frawa: This was an attempt to enable applying IntelliJ code style without installing Idea.
Do you have a better approach using VS Code, or Eclipse, or command line?

mathieucarbou: No... Vscode is a hell if you were used to intellij.
We are trying it in our team - we were all using intellij, sadly we cannot anymore.
Although I was able to find equivalent things for basic features, for more advanced ones and formatting, I didn't yet.

mathieucarbou: Clang works fine for c++, I don't know if ther is an equivalent for java

mathieucarbou:
@frawa also do not forget to rebase;-)

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/GitLookup.java
frawa: Done

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
frawa: `ForwardingMap` it is.

Including `file.name` with forwarding will be tricky, as it is part of the read only props passed to the adjusting call.
You'd end up with a deeply nested structure of forwarding maps, while accumulating resolver adjustments. (I tried that earlier and did not like it.)

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightRangeProvider.java
frawa: Unfortunately, Java's missing approach to checked exceptions for Lambdas prevents this slick solution.
I went with an atomic ref and a helper function to avoid code duplication.

mathieucarbou: There is a way to hack around that using templates for exceptions but this is a hacky way lol ;-) but since not a public API would be acceptable. The atomic ref is fine also ;-)

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
mathieucarbou: file.name could be hard coded or put in a map copied from the global props.

mathieucarbou:
@frawa thanks! Will check the PR again today.
Can you please also squash all the commits into one ?

mathieucarbou:
@frawa : checked the PR. 
Introducing this idea of forwarding map now makes me think that we do not need all these supplier maps at all but only 1 new map implementation like a `LazyMap` would probably be cleaner. It allows to pass suppliers (memoized or not) with not touching the PropertiesProvider interface.

I will push a commit for that this afternoon to cleanup: using a custom map implementation will allow to leave a lot of the API in-place.

frawa:
> @frawa : checked the PR. Introducing this idea of forwarding map now makes me think that we do not need all these supplier maps at all but only 1 new map implementation like a `LazyMap` would probably be cleaner. It allows to pass suppliers (memoized or not) with not touching the PropertiesProvider interface.
> 
> I will push a commit for that this afternoon to cleanup: using a custom map implementation will allow to leave a lot of the API in-place.

That was my very first approach, see [this commit](https://github.com/frawa/license-maven-plugin/pull/1/commits/3751bdcb8e802f0f7bc3b947add6940466f54c7e#diff- eb856076d9adf427c53efe7f9f39b4da7280234b5320d08568f3c8b66c68191e), but I found it too much work to make `LazyMap` safe to use. 

The supplier map seems to convey the intent in a clearer way, IMHO.

I'll let you judge and take over to you liking.

mathieucarbou:
> > @frawa : checked the PR. Introducing this idea of forwarding map now makes me think that we do not need all these supplier maps at all but only 1 new map implementation like a `LazyMap` would probably be cleaner. It allows to pass suppliers (memoized or not) with not touching the PropertiesProvider interface.
> > I will push a commit for that this afternoon to cleanup: using a custom map implementation will allow to leave a lot of the API in-place.
> 
> That was my very first approach, see [this commit](https://github.com/frawa/license-maven-plugin/pull/1/commits/3751bdcb8e802f0f7bc3b947add6940466f54c7e#diff-eb856076d9adf427c53efe7f9f39b4da7280234b5320d08568f3c8b66c68191e), but I found it too much work to make `LazyMap` safe to use.
> 
> The supplier map seems to convey the intent in a clearer way, IMHO.
> 
> I'll let you judge and take over to you liking.

Yes more work to create this map, but less after. It has to be done anyway because this is not possible to modify the `PropertiesProvider` since this is public API to allow SPI extensions.

mathieucarbou:
@frawa : I have squashed your original commits into one to keep your work as-is, and I added a second commit to update this PR. 

I have introduced a `LazyMap`, supplier based, in order not to change the public API of `PropertiesProvider`, this lazy map being able to alternatively grab keys from a default supplier function (which is our global properties map).

Would you be able to test these changes to ensure they are still performant ?

I will wait for your feedback before merging.

Thanks!

Thanks!

frawa:
> @frawa : I have squashed your original commits into one to keep your work as-is, and I added a second commit to update this PR.
> 
> I have introduced a `LazyMap`, supplier based, in order not to change the public API of `PropertiesProvider`, this lazy map being able to alternatively grab keys from a default supplier function (which is our global properties map).
> 
> Only 3 files are updated following this change.
> 
> Would you be able to test these changes to ensure they are still performant ?
> 
> I will wait for your feedback before merging.
> 
> Thanks!

Nice.
Confirmed, the performance on the repo I am looking at is still improved by at least a factor 3.

hazendaz:
Thanks @frawa I've also seen same results on mybatis project.  Using a profiler, here are my results.

Using currently 4.6 release
<html><body>

com.mycila:license-maven-plugin:4.6:format {execution: default} | 79.60 s
-- | --



</body>
</html>

Using current master here (and a very few extra cleanup items I have - nothing impactful)

<html><body>

om.mycila:license-maven-plugin:5.0.0-SNAPSHOT:format {execution: default} | 24.23 s
-- | --



</body>
</html>

I repeated the formatting and touched a few files to make sure they are working as expected with same results.

In the overall project for mybatis I tested it on, its no longer the worst performing aspect.  Its not nearly as fast as general formatters for obvious reasons but huge improvement non the less.

Issue #887, opened 2025-01-15
Introduce workspace parameter


mathieu:
> @mathieucarbou , I'll need some more context on your vision of the `<workspace>`. What does it do compared to LicenceSet's `<includes>` and `<excludes>` ?

`workspace` would define the global workspace to work one

- workspace.baseDir == project root directory
- workspace.includes defaulted to null
- workspace.excludes defaulted to null

This concept exists already but is implicit: 
- the workspace.baseDir is currently `defaultBasedir`, 
- we do not constrain to any include (the whole workspace is visible) 
- and we do not constrain to some excludes (nothing is excluded by default).

The idea is not make it explicitly configurable.

`LicenceSet` are used to "split" the workspace into several groups of file and assigning them a license. So they inherit the workspace baseDir, includes, excludes. licensSet.baseDir can be overridden but must be a sub directory of the workspace (@hazendaz : I don't know if we make sure of that somewhere, but with the workspace concept we will be)

As I see things for includes/excludes:

- if workspace.includes is defined in POM, then it will be added to all licenseSet.includes
- if workspace.excludes is defined in POM, then it will be added to all licenseSet.excludes
- if workspace.baseDir is defined in POM, then it will be used to set licenseSet.baseDir (if not set) - like what happens now with defaultBasedir, and if licenseSet.baseDir is set, we have to check that it is same or a subfolder of workspace.baseDir.

For backward compat', `defaultBasedir` must be marked deprecated, without any default now ,and if set by the user, its value will fill "workspace.baseDir". The property `license.basedir` should be moved to `workspace.baseDir`

In the case of a commit hook, then the only property to add would be:

`-D license.workspace.includes=<list of file paths>`

This behind will be transferred to licenseset.includes and will have at the end the same effect of my comment in this PR: https://github.com/mathieucarbou/license-maven-plugin/pull/878#issuecomment-2585331008

Also, this "design" does not change the scanner logic, which is still taking place but based on a refinement of the baseDir / includes / excludes.

@mathieucarbou @hazendaz @frawa : let me know what you think of that ;-)

_Originally posted by @mathieucarbou in https://github.com/mathieucarbou/license-maven-plugin/issues/876#issuecomment-2586659052_

mathieu:
@mathieucarbou, I've an issue with the following requirement
[...]

PR #888, opened 2025-01-15
feat: #887 Added WorkSpace class

mathieu:


license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/util/FileUtils.java
```diff
@@ -133,4 +133,10 @@ public static void copyFilesToFolder(File src, File dst) {
       }
     });
   }
+
+  public static boolean isSubfolder(File subfolder, File folder) {
+    String subfolderPath = subfolder.getAbsolutePath();
+    String folderPath = folder.getAbsolutePath();
+    return subfolderPath.startsWith(folderPath) && subfolderPath.length() > folderPath.length() && subfolderPath.charAt(folderPath.length()) == File.separatorChar;
+  }
```
mathieucarbou, line 141: This could be moved to a one-liner with the java Path api. 
Any reason to not use it ?

license-maven-plugin-git/src/main/java/com/mycila/maven/plugin/license/git/CopyrightAuthorProvider.java
```diff
@@ -40,7 +40,7 @@ public class CopyrightAuthorProvider implements PropertiesProvider {
 
   @Override
   public void init(AbstractLicenseMojo mojo, Map<String, String> currentProperties) {
-    gitLookup = GitLookup.create(mojo.defaultBasedir, currentProperties);
+    gitLookup = GitLookup.create(mojo.legacyDefaultBasedir, currentProperties);
```
mathieucarbou, line 43: I would use mojo.workspace.baseDir for all these occurences to avoid using deprecated fields in the code base

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -94,9 +99,12 @@ public abstract class AbstractLicenseMojo extends AbstractMojo {
    * This is named `defaultBaseDirectory` as it will be used as the default
    * value for the base directory. This default value can be overridden
    * in each LicenseSet by setting {@link LicenseSet#basedir}.
+   *
+   * @deprecated use {@link WorkSpace#basedir}
    */
+  @Deprecated
   @Parameter(property = "license.basedir", defaultValue = "${project.basedir}", alias = "basedir", required = true)
-  public File defaultBasedir;
+  public File legacyDefaultBasedir;
```
mathieucarbou, line 107: I would make it optional and null by default - in order to catch users using it explicitely.
`@Parameter(property = "license.basedir", alias = "basedir")`

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -492,13 +500,29 @@ protected final void execute(final Callback callback) throws MojoExecutionExcept
       }
 
       // make default base dir canonical
-      this.defaultBasedir = this.getCanonicalFile(this.defaultBasedir, "license.basedir");
+      workspace.basedir = getCanonicalFile(firstNonNull(workspace.basedir, legacyDefaultBasedir), "license.workspace.basedir");
```
mathieucarbou, line 503: I would set `workspace.basedir` like this:

`workspace.basedir = getCanonicalFile(workspace.basedir, "license.workspace.basedir");`

but then `if(workspace.basedir == null)`, I would look for the `legacyDefaultBasedir`. And if this one is null, I would use the project.basedir.

The idea is to make the new config always a priority over the legacy one, the legacy one being used only as a fallback.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/WorkSpace.java
```diff
@@ -0,0 +1,39 @@
...
+public class WorkSpace {
+
+  /**
+   * The base directory, in which to search for project files.
+   * <p>
+   * It will be used as the default value for the base directory of each LicenseSet.
+   * This default value can be overridden in each LicenseSet by setting {@link LicenseSet#basedir}.
+   */
+
+    @Parameter(property = "license.workspace.basedir", defaultValue = "${project.basedir}", alias = "basedir")
```
mathieucarbou, line 31: I wouldn't set any default to handle backward compatibility, otherwise we cannot know if it was explicitly set by the user of set by maven

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -492,13 +500,29 @@ protected final void execute(final Callback callback) throws MojoExecutionExcept
...
 
       // collect all the license sets together
       final LicenseSet[] allLicenseSets;
 
       // if we abandon the legacy config this contiguous block can be removed
       final LicenseSet legacyLicenseSet = convertLegacyConfigToLicenseSet();
+
+      if (workspace.basedir != null) {
```
mathieucarbou, line 511: this should never happen: the code above ensures `workspace.basedir` will be always set. We could even have a `requireNonNull()` check to make sure of that, to help for tests.

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -492,13 +500,29 @@ protected final void execute(final Callback callback) throws MojoExecutionExcept
...
+        if (legacyLicenseSet != null && legacyLicenseSet.basedir != null) {
+          if (!FileUtils.isSubfolder(legacyLicenseSet.basedir, workspace.basedir)) {
+            throw new MojoExecutionException("Legacy basedir parameter should be a subfolder of the workspace basedir.");
```
mathieucarbou, line 514: The logic is to fail if a licenseSet baseDir is outside the workspace. 

It is perfectly legal to have only one licenset in a project with its baseDir set to the same value of the workspace baseDir

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -492,13 +500,29 @@ protected final void execute(final Callback callback) throws MojoExecutionExcept
...
+          }
+        }
+        for (LicenseSet licenseSet : licenseSets) {
+          if (licenseSet.basedir != null) {
+            if (!FileUtils.isSubfolder(licenseSet.basedir, workspace.basedir)) {
+              throw new MojoExecutionException(String.format("LicenseSet basedir parameter [%s] should be a subfolder of the workspace basedir.", licenseSet.basedir.getPath()));
```
mathieucarbou, line 520: same comment

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -671,7 +697,7 @@ private void executeForLicenseSet(final LicenseSet licenseSet, final Callback ca
       };
 
       final DocumentFactory documentFactory = new DocumentFactory(
-          firstNonNull(licenseSet.basedir, defaultBasedir), buildMapping(),
+          firstNonNull(licenseSet.basedir, workspace.basedir), buildMapping(),
```
mathieucarbou, line 700: same simplification: we can only use `licenseSet.basedir`

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/AbstractLicenseMojo.java
```diff
@@ -786,9 +812,9 @@ private Map<String, String> getDefaultProperties() {
   private String[] listSelectedFiles(final LicenseSet licenseSet) {
     final boolean useDefaultExcludes = (licenseSet.useDefaultExcludes != null ? licenseSet.useDefaultExcludes : defaultUseDefaultExcludes);
     final Selection selection = new Selection(
-        firstNonNull(licenseSet.basedir, defaultBasedir), licenseSet.includes, buildExcludes(licenseSet), useDefaultExcludes,
+        firstNonNull(licenseSet.basedir, workspace.basedir), licenseSet.includes, buildExcludes(licenseSet), useDefaultExcludes,
         getLog());
-    debug("From: %s", firstNonNull(licenseSet.basedir, defaultBasedir));
+    debug("From: %s", firstNonNull(licenseSet.basedir, workspace.basedir));
```
mathieucarbou, line 817: same simplification: we can only use `licenseSet.basedir`

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/WorkSpace.java
```diff
@@ -0,0 +1,39 @@
...
+    File basedir;
+
+    @Parameter
+    String[] includes = new String[0];
+
+    @Parameter
+    String[] excludes = new String[0];
```
mathieucarbou, line 38: I didn't see any use of that ? Maybe because the PR is still in draft ?

mathieucarbou (commented):
some preliminary comments ;-)
thanks a lot for tackling this refactoring!

license-maven-plugin/src/main/java/com/mycila/maven/plugin/license/WorkSpace.java
mathieu: Yes this is to be done after I get the gist of the workspace introduction done.

PR #891, opened 2025-01-20
Build cleanup items

hazendaz:
- Added mockito bom to ensure its usage of byte buddy doesn't get manipulated by another other libs so higher jdks work without issues.
- Drop jdk 22 on build matrix
- Add jdk 23 on build matrix
- Add jdk 24-ea on build matrix
- Do not care about byte compatibility for 'test' or 'provided' scope as neither affects consumers
- Fix javadoc link to java 11 fixing jdk 23+

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        java: [11, 17, 21, 22]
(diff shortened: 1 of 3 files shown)


hazendaz:
note: mockito is not a direct dependency but is used through maven plugin testing harness.  Because maven entirely dropped maven 3 support in that stream after the version we are on, we are stuck, so this corrects the stack in most simplistic way.

mathieucarbou:
@hazendaz : thanks a LOT for going through all the maintenance of this project over the past years! you do an amazing job to keep it in a clean state. That's really awesome!

PR #896, opened 2025-01-26
[eol] Fix end of file markers

hazendaz:
ok this is the remainder of the files that were in normal code space that did not end with end of file marker per git standards.  As you can see, github does a nice job marking the change ;)


hazendaz:
need to work on tests, will come back on this.

PR #897, opened 2025-01-26
[tests] Use path as temporary directory instead of file which we only…

hazendaz:
… grabbed path from

### DIFF
--- a/license-maven-plugin-fs/src/test/java/com/mycila/maven/plugin/license/fs/CopyrightRangeProviderTest.java
+++ b/license-maven-plugin-fs/src/test/java/com/mycila/maven/plugin/license/fs/CopyrightRangeProviderTest.java
@@ -77,13 +77,13 @@ class CopyrightRangeProviderTest {
   }
 
   private static Document newDocument(String relativePath) {
-    Path path = Path.of(fsRepoRoot.toString(), relativePath.replace('/', File.separatorChar));
+    Path path = fsRepoRoot.resolve(relativePath.replace('/', File.separatorChar));
     return new Document(path.toFile(), null, StandardCharsets.UTF_8, new String[0], null);
   }
 
   @BeforeAll
   static void beforeClass() throws IOException {
-    fsRepoRoot = Path.of(tempFolder.toString(), "fs-test-repo");
+    fsRepoRoot = tempFolder.resolve("fs-test-repo");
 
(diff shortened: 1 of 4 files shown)


license-maven-plugin-fs/src/test/java/com/mycila/maven/plugin/license/fs/CopyrightRangeProviderTest.java
```diff
@@ -84,7 +84,7 @@ private static Document newDocument(String relativePath) {
 
   @BeforeAll
   static void beforeClass() throws IOException {
-    fsRepoRoot = Path.of(tempFolder.toPath() + File.separator + "fs-test-repo");
+    fsRepoRoot = Path.of(tempFolder + File.separator + "fs-test-repo");
```
mathieucarbou, line 87: In theory this should be: `Path.of(tempFolder, "fs-test-repo");`

should this (and the others) be changed also in this PR ?

mathieucarbou (commented):
a little comment ;-)

license-maven-plugin-fs/src/test/java/com/mycila/maven/plugin/license/fs/CopyrightRangeProviderTest.java
mathieucarbou: ah! just saw tempFolder is a Path already... so: tempFolder.resolve(...) ?

hazendaz: makes sense, yes will change that.

hazendaz: Used toString(), haven't looked at resolve yet.  There were others. I can look at that if this still looks to need some more work.

mathieucarbou: Yeah it does a lot of operations for nothing ;-)

fsRepoRoot = tempFolder.resove("fs-test-repo")

hazendaz: pushed updates

mathieucarbou (approved):

Issue #970, opened 2025-09-10
Support for more licenses?


luca-vercelli:
**Describe the feature**

Please consider adding headers for following common open source licenses:

* Eclipse's EPL (v.1 and 2) 
* European EUPL (v. 1.1 and 1.2).

Some of them are included e.g. in https://github.com/mojohaus/license-maven-plugin/tree/master/src/main/resources/META-INF/licenses

**Any alternative you've considered ?**

* include license header files in every project
* change license plugin

If there is some interest in these features, I could provide a PR.

hazendaz:
Yes please submit a PR

Sent from my Verizon, Samsung Galaxy smartphone
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________
From: luca-vercelli ***@***.***>
Sent: Wednesday, September 10, 2025 5:27:17 AM
To: mathieucarbou/license-maven-plugin ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [mathieucarbou/license-maven-plugin] Support for more licenses? (Issue #970)

[https://avatars.githubusercontent.com/u/9864296?s=20&v=4]luca-vercelli created an issue (mathieucarbou/license-maven-plugin#970)<https://github.com/mathieucarbou/license-maven-plugin/issues/970>

Describe the feature

Please consider adding headers for following common open source licenses:

  *   Eclipse's EPL (v.1 and 2)
  *   European EUPL (v. 1.1 and 1.2).

Some of them are included e.g. in https://github.com/mojohaus/license-maven-plugin/tree/master/src/main/resources/META-INF/licenses

Any alternative you've considered ?

  *   include license header files in every project
  *   change license plugin

If there is some interest in these features, I could provide a PR.

hazendaz:
@luca-vercelli See #1029 , that work for you?

PR #974, opened 2025-09-18
support for EPL and EUPL

luca-vercelli:
As described in issue #970.
Added header files for EPL 1.0, EPL 2.0, EUPL 1.x


mathieucarbou (approved):
@luca-vercelli : you'll probably need to update some unit tests...

luca-vercelli:
Tests do not fail. Do you want that I add some new test? If so, where?

hazendaz:
Thanks @luca-vercelli.  Its been a little while since we released.  I have a few things I need to get over to repo for global ignores and possible to get checkstyle updated and few others.  I'm going to try to work those out in next week or so then suggest we move for release so we pick this support up.  If you see other items that might be useful but missing, feel free to send more PRs.

PR #990, opened 2025-10-19
Dependency adjustments

hazendaz:
- For commons io and commons pool2, those are coming into runtime.  In case of commons io, its really old due to resolution issues and for commons pool2, that comes via the xml code this project has that hasn't been updated in something like 14 years.  Added into dependency management to ensure when we are providing those public facing items, we are providing secure and maintained versions.  This is especially important as some cve's have been flagged recently on commons modules.

- Maven finally backported to maven 3 for the testing harness so that is applied, I initially was adding additional overrides but have removed the guice override I had earlier as that is now fixed.

- For dependabot, I added snippet to it so it starts updating actions.  I had been relying on my usage of my forked Renovate usage to do that so this will set it correctly for this repo, I did update the latest action update already here as well.

- For github actions, I updated the jdks involved to match current expectations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,3 @@ Often necessary to exclude a number of files that would otherwise grow the exclu
 Version >= 5.0.0 will require Java 11.
 
 Version >= 6.0.0 will require Java 17.
-
-### Supported by
-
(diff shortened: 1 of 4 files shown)


mathieucarbou (approved):

mathieucarbou:
@hazendaz sorry I didn't see all the pings... I was quite over busy the past weeks.

hazendaz:
@mathieucarbou No problem!  Thanks.

